PR #9, opened 2020-09-29
Refactor/revisions

HarmdR:


revisions.py
```diff
@@ -0,0 +1,476 @@
...
+    if case == 'A':
+        return [
+            [rev_hours(6), 1, 1],
+            [rev_hours(12), 1, 0],
+            [rev_hours(18), 1, 0],
+            [rev_days(1), 1, 0],
+            [rev_days(2), 1, 0],
+            [rev_days(3), 1, 0],
+            [rev_days(4), 1, 0],
+            [rev_days(6), 1, 0],
```
lwesterhof: These two lines should be swapped I think, 5 days is after 6 days.

revisions.py
```diff
@@ -0,0 +1,476 @@
...
+            return False
+
+    log.write(ctx, "revision_remove('" + revision_id + "'): Revision ID not found or permission denied.")
+    return False
+
+
+""" Helper functions for easy time conversion eventually in seconds """
+
+
+def rev_minutes(minutes):
```
lwesterhof: It may be cleaner to define constants instead of these helper functions.

Something like:
`HOUR = 3600`
`DAY = 86400`
`WEEK = 604800`

And then define the buckets as HOUR * 6, HOUR * 12, HOUR * 18, DAY,  DAY * 2, etc.

revisions.py
```diff
@@ -0,0 +1,476 @@
...
+                while count < nr_to_be_removed:
+                    deletion_candidates.append(rev_list[len(rev_list) + (bucket_start_index) - count])
+                    count += 1
+
+        bucket_counter += 1  # To keep conciding with strategy list
+
+    return deletion_candidates
+
+
+def calculate_end_of_calendar_day(ctx):
```
lwesterhof, line 256: We don't have to use msi.get_icat_time() here. msi.get_icat_time() returns the system time for the iCAT server, so we can get that directly from Python because we are on the iCAT server.

Something like this should work:
```
# Get datetime of tomorrow.
tomorrow = datetime.date.today() + datetime.timedelta(1)

# Convert tomorrow to unix timestamp.
unix_time = tomorrow.strftime("%s")

return unix_time
```

lwesterhof (changes requested):

revisions.py
HarmdR: Done

revisions.py
HarmdR: DOne

revisions.py
HarmdR: Done

revisions.py
```diff
@@ -0,0 +1,482 @@
...
+        msi.set_acl(ctx, "recursive", "admin:own", user.full_name(ctx), revision_store)
+        msi.set_acl(ctx, "recursive", "inherit", user.full_name(ctx), revision_store)
+
+    end_of_calender_day = int(endOfCalendarDay)
+    if end_of_calender_day == 0:
+        end_of_calendar_day = calculate_end_of_calendar_day(ctx)
+
+    # get definition of buckets
+    buckets = revision_bucket_list(ctx, bucketcase)
+    log.write(ctx, buckets)
```
lwesterhof: There is still some debug logging left. To prevent cluttering the logs we can remove these or change them to log.debug if they are valuable. Log.debug statements will only show on development instances.

revisions.py
```diff
@@ -0,0 +1,482 @@
...
+                       "AND COLL_NAME like '" + startpath + "%' ",
+                       offset=0, limit=None, output=query.AS_DICT)
+
+    # qtotalrows.total_rows() moet worden verminderd met het aantal ontdubbelde entries
+    return {'total': qtotalrows.total_rows() - multiple_counted,
+            'items': revisions}
+
+
+@api.make()
+def api_revisions_list(ctx, path):
```
lwesterhof, line 367: API functions should always be documented, because this is used to generate the API documentation. We use the Sphinx docstring format: https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html For examples see https://github.com/UtrechtUniversity/irods-ruleset-uu/blob/8a31c9d2aa98ace579abd47f76211e70208d40b3/folder.py#L102

lwesterhof (changes requested):

HarmdR:
Changes added 👍

lwesterhof (approved):

tests/features/api_revisions.feature
```diff
@@ -0,0 +1,33 @@
...
+        Given the Yoda revision API is queried with "<path>"
+	    Then the response status code is "200"
+	    And revisions list is found 
+
+        Examples:
+        | user       | path                                                                  |
+        | researcher | /tempZone/home/research-initial/testdata/SIPI_Jelly_Beans_4.1.07.tiff |
+
+
+    Scenario: Restore a revision
```
lwesterhof, line 25: This scenario fails because the revision id changes.
Maybe we can add a given step 'And revision exists in "<path>"' that retrieves the revision id of an existing revision. Then in the API request step this revision id can be used. 

For examples see datarequest API test: https://github.com/UtrechtUniversity/irods-ruleset-uu/blob/4777f16238ff8377d5e0b2c73bc1baa38acb7850/tests/features/api_datarequest.feature#L15

lwesterhof (changes requested):

PR #14, opened 2020-10-07
Use mail util for mails

aucuparia:


datarequest.py
```diff
@@ -1349,10 +1259,377 @@ def mail_datarequest_submitted(ctx, researcher_email, researcher_name, request_i
...
+
+With kind regards,
+YOUth
+""".format(request_id, request_id))
+
+
+def mail_datarequest_evaluation_submit_researcher(ctx, decision, researcher_email, researcher_name,
+                                                  request_id, datamanager_email = None,
+                                                  feedback_for_researcher = None):
+    if decision == "Approved":
```
lwesterhof: I think it is cleaner to have this logic outside the mail template. Make separate mail templates and move the logic one level higher.

datarequest.py
```diff
@@ -1021,31 +956,24 @@ def api_datarequest_review_submit(ctx, data, request_id):
...
-            value = row["META_DATA_ATTR_VALUE"]
-            if name == "name":
-                researcher_name = value
-            elif name == "email":
-                researcher_email = value
+        datarequest = jsonutil.read(ctx, coll_path + "/datarequest.json")
+
+        researcher = datarequest['researchers']['contacts'][0]
+
+        researcher_name = researcher['name']
```
lwesterhof: Is there any reason to not use researcher['name'] directly? They are only used in the mail call? Same for researcher_email.

lwesterhof (changes requested):

datarequest.py
aucuparia: Agreed. Fixed.

datarequest.py
aucuparia: No. Fixed.

lwesterhof (approved):

PR #20, opened 2020-10-12
Yda 3731

reavdhoef:
Refactored the datarequest_is_owner function + its calls.

### DIFF
--- a/datarequest.py
+++ b/datarequest.py
@@ -586,7 +586,7 @@ def api_datarequest_is_owner(ctx, request_id, user_name):
        :rtype: bool
     """
 
-    is_owner = False;
+    is_owner = False
 
     try:
         is_owner = datarequest_is_owner(ctx, request_id, user_name)
--- a/datarequest.py
+++ b/datarequest.py
@@ -582,11 +582,18 @@ def api_datarequest_is_owner(ctx, request_id, user_name):
        :param user_name: Username of the user whose ownership is checked
        :type user_name: str
 
-       :raises Exception: It was not possible to unambiguously determine the owner of the data request (either 0 or > 1 results for the data request)
        :return: `True` if ``user_name`` matches that of the owner of the data request with id ``request_id``, `False` otherwise
        :rtype: bool
     """
-    return datarequest_is_owner(ctx, request_id, user_name)
+
+    is_owner = False;
+
+    try:
+        is_owner = datarequest_is_owner(ctx, request_id, user_name)
+    except Exception as e:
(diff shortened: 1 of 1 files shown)


datarequest.py
```diff
@@ -573,24 +573,33 @@ def api_datarequest_datamanager_review_get(ctx, request_id):
...
+       :type request_id: str
+       :param user_name: Username of the user whose ownership is checked
+       :type user_name: str
 
-    return result['owner']
+       :raises Exception: It was not possible to unambiguously determine the owner of the data request (either 0 or > 1 results for the data request)
+       :return: `True` if ``user_name`` matches that of the owner of the data request with id ``request_id``, `False` otherwise
+       :rtype: bool
+    """
+    return datarequest_is_owner(ctx, request_id, user_name)
```
lwesterhof: Can you catch the Exception here and return some extra information with api.Error. For example: https://github.com/UtrechtUniversity/irods-ruleset-uu/blob/f6c7a6de6f5aa4c22b6a4b6e1a5a89916d57e1bb/datarequest.py#L199

lwesterhof (changes requested):

datarequest.py
reavdhoef: Updated. I also removed the "raises Exception" line from the documentation; is this how it should be documented?

datarequest.py
```diff
@@ -573,24 +573,40 @@ def api_datarequest_datamanager_review_get(ctx, request_id):
...
+       :rtype: bool
+    """
+
+    is_owner = False
 
-    if not (result['status'] == 0):
-        return api.Error("error", "Something went wrong in determining datarequest ownership")
+    try:
+        is_owner = datarequest_is_owner(ctx, request_id, user_name)
+    except Exception as e:
```
lwesterhof: Better to raise and catch an error.UUError. Otherwise other bugs are also catched by this Exception.

When another Exception occurs it will be catched higher in de API layer.

lwesterhof (changes requested):

datarequest.py
lwesterhof: And if you raise the Exception with a message you can use it here as e.message. If you don't use it you can remove "as e".

datarequest.py
lwesterhof: Yes, LGTM

datarequest.py
reavdhoef: Updated: now creating error.UUError instead of generic Exception.

lwesterhof (approved):

PR #22, opened 2020-10-13
Status transition checks

aucuparia:


lwesterhof (approved):
LGTM. To be fixed later, not all status transitions are covered right now. Maybe better to integrate status_transition_allowed into status_set? Then you can't miss any transitions and status_set can throw and error if it is invalid.

aucuparia:
> LGTM. To be fixed later, not all status transitions are covered right now. Maybe better to integrate status_transition_allowed into status_set? Then you can't miss any transitions and status_set can throw and error if it is invalid.

All status transitions that a user can execute through Yoda are covered, but manually changing the `status` metadata field is still possible. An issue has been added to the backlog to implement status transition checks as a policy on the `status` metadata field.

PR #27, opened 2020-10-20
Yda 3662 - Mail test functionality

reavdhoef:
Things to consider:

1. in the ticket it was specified to have some information in the sent email about which instance is sending the email. From my understanding I could _maybe_ find out the name given to the portal, but that is not by definition identifying. Since the user executing the test has to pass the target email address the user can know for sure that they sent the email and they then know/can find out what instance they are on.

2. This should only be executable by admins. I purposely left the permissions for the script the same as those of others in the tools folder, since all those scripts should dnot be accessible for regular users. Is it fair to assume this is safe because users cannot login directly to the machine and they also cannot execute the rules related to them? Or should permissions be said explicitly as well?

### DIFF
--- a/mail.py
+++ b/mail.py
@@ -12,7 +12,8 @@ from email.mime.text import MIMEText
 from util import *
 
 __all__ = ['rule_mail_new_package_published',
-           'rule_mail_your_package_published']
+           'rule_mail_your_package_published',
+           'rule_mail_test']
 
 
 def send(ctx, to, actor, subject, body):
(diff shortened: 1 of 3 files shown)


lwesterhof (approved):

lwesterhof:
1. Agree

2. The script should not be accessible for regular users. However, the rule is accessible for all users. The safety check is in the in the mail send function, there it checks if the caller is a rodsadmin.

PR #30, opened 2020-10-22
3737 validate against json schema

aucuparia:


datarequest.py
```diff
@@ -37,20 +38,26 @@
...
 
-SCHEMA_FILENAME   = "schema.json"
-UISCHEMA_FILENAME = "uischema.json"
-DR_FILENAME       = "datarequest.json"
-PR_REV_FILENAME   = "preliminary_review.json"
-DM_REV_FILENAME   = "datamanager_review.json"
-EVAL_FILENAME     = "evaluation.json"
-ASSIGN_FILENAME   = "assignment.json"
+DRCOLLECTION      = "home/datarequests-research"
+DR                = "datarequest"
```
lwesterhof: Instead of defining every filename, we could define FILENAME_EXT = ".json" once and use this.

datarequest.py
```diff
@@ -37,20 +38,26 @@
...
+PR_REV            = "preliminary_review"
+DM_REV            = "datamanager_review"
+REV               = "review"
+ASSIGN            = "assignment"
+EVAL              = "evaluation"
```
lwesterhof: EVAL -> EVALUATION
REV -> REVIEW
etc.

No need to use abbreviation. I would read REV as revision

util/pathutil.py
```diff
@@ -107,11 +107,11 @@ def test(r, space):
 
     return (namedtuple('PathInfo', 'space zone group subpath'.split())
```
lwesterhof: This change is not needed?

datarequest.py
```diff
@@ -273,6 +304,10 @@ def api_datarequest_submit(ctx, data, previous_request_id):
     coll_path = "/{}/{}/{}".format(user.zone(ctx), DRCOLLECTION, request_id)
     file_path = "{}/{}".format(coll_path, DR_FILENAME)
 
+    # Validate data against schema
+    if not datarequest_data_valid(ctx, data, DR):
+        return api.Error("validation_fail", "Data did not pass validation.")
```
lwesterhof: Specify what data could not pass validation?

Validation of datarequest failed?

datarequest.py
```diff
@@ -240,10 +247,14 @@ def transform_title(row):
...
 def api_datarequest_schema_get(ctx, schema_name):
+    return datarequest_schema_get(ctx, schema_name)
+
+
+def datarequest_schema_get(ctx, schema_name):
     """Get schema and uischema of a datarequest form
 
        Arguments:
-       schema_name -- Name of schema.
+       schema_name -- Name of schema
```
lwesterhof, line 253: Sphinx docstyle: https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html

```suggestion
       :param schema_name: Name of JSON schema
```

lwesterhof (changes requested):

util/pathutil.py
aucuparia: Code looks fine, yes. This only fixes a flake8 error. Revert?

lwesterhof: Should not be an error, passes in development?

aucuparia: Probably due to running older flake8/pyflakes versions. Reverted the commit and pushed a new one.

lwesterhof (approved):

PR #33, opened 2020-10-23
Replace api_group_user_exists with more restricted API functions

aucuparia:


datarequest.py
```diff
@@ -757,6 +760,48 @@ def datarequest_is_reviewer(ctx, request_id):
     return is_reviewer
 
 
+@api.make()
+def api_datarequest_is_bod_member(ctx):
+    """
+    Check if given user is BOD member
+
+    :param user: User name
```
lwesterhof: There is no parameter user.

lwesterhof (changes requested):

lwesterhof (approved):

PR #49, opened 2021-01-13
Set default for Data_Type to Datapackage.

lwesterhof:
Default is Dataset, this translates to Datapackage in form.


reavdhoef (approved):
LGTM

PR #53, opened 2021-02-09
if  resource_replica is not defined (empty string) ignore uuReplicate…

stefan-wolfsheimer:
…Asynchronously


lwesterhof:
Thanks!

PR #63, opened 2021-03-15
API for saving and loading settings

lwesterhof:
Uses the sudo policies to allow setting a AVU as a normal rodsuser.

### DIFF
--- a/settings.py
+++ b/settings.py
@@ -5,9 +5,12 @@ __copyright__ = 'Copyright (c) 2021, Utrecht University'
 __license__   = 'GPLv3, see LICENSE'
 
 from util import *
+from util.query import Query
 
 __all__ = ['api_settings_load',
            'api_settings_save']
+
+USER_SETTINGS = {"mail_notifications": {"default": True, "values": [True, False]}}
 
 
 @api.make()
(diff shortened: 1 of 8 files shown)


HarmdR:
Ziet er prima uit 👍

settings.py
```diff
@@ -0,0 +1,54 @@
...
+    :param settings: Dict with settings to be saved
+
+    :returns: API status
+    """
+    for a, v in settings.items():
+        if a in USER_SETTINGS and v in USER_SETTINGS[a]["values"]:
+            try:
+                ctx.uuUserModify(user.full_name(ctx), "{}{}".format(SETTINGS_KEY, a), str(v), '', '')
+            except Exception as e:
+                return api.Error('internal', 'Saving settings failed. Please try again')
```
reavdhoef: Does this error message end up at the user? If so, I would suggest to change the error message to something along the lines of "Saving settings failed. Please try again shortly. If the problem persists after a few tries, please contact Yoda FB."

reavdhoef (approved):
Small remark on error message, but looks good.

HarmdR (approved):

settings.py
lwesterhof: Thanks, done.

PR #68, opened 2021-04-06
Add support for user notifications

lwesterhof:


notifications.py
```diff
@@ -0,0 +1,78 @@
...
+                             "USER_NAME = '{}' AND USER_TYPE = 'rodsuser' AND META_USER_ATTR_NAME like '{}_%%'".format(user.name(ctx), NOTIFICATION_KEY))]
+
+    notifications = []
+    for result in results:
+        try:
+            notification = jsonutil.parse(result)
+            notification["datetime"] = (datetime.fromtimestamp(notification["timestamp"])).strftime('%Y-%m-%d %H:%M')
+            notifications.append(notification)
+        except Exception:
+            continue
```
reavdhoef, line 64: What scenario can cause this? Should it have some logging?

reavdhoef (approved):
Small remark on error handling. For the rest, LGTM

notifications.py
lwesterhof: In this scenario the notification (stored as JSON) can not be parsed. Will add some logging after merge.

PR #70, opened 2021-04-08
Refactor group data API

lwesterhof:
- Combine group_data and group_data_filtered APIs
- Return group data directly as hierarchy to be used in frontend


reavdhoef (approved):
LGTM

PR #75, opened 2021-05-11
Pyportal intake additions

HarmdR:
He Lazlo,
Wat kleine veranderingen ten behoeve van de intake module

Gr,
Harm

### DIFF
--- a/intake_dataset.py
+++ b/intake_dataset.py
@@ -52,9 +52,6 @@ def intake_report_export_study_data(ctx, study_id):
                                            "COLL_NAME like '{}%'".format(set_path),
                                            genquery.AS_LIST, ctx)
             for row in result:
-                log.write(ctx, 'export')
-                log.write(ctx, row[0])
-                log.write(ctx, row[1])
                 real_datasets[set_path]['totalFiles'] = int(row[0]) / 2
                 totalFileSize = 0
                 if row[1]:
--- a/intake_dataset.py
+++ b/intake_dataset.py
@@ -52,8 +52,14 @@ def intake_report_export_study_data(ctx, study_id):
                                            "COLL_NAME like '{}%'".format(set_path),
                                            genquery.AS_LIST, ctx)
             for row in result:
+                log.write(ctx, 'export')
+                log.write(ctx, row[0])
+                log.write(ctx, row[1])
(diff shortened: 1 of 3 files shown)


intake_dataset.py
```diff
@@ -52,8 +52,14 @@ def intake_report_export_study_data(ctx, study_id):
                                            "COLL_NAME like '{}%'".format(set_path),
                                            genquery.AS_LIST, ctx)
             for row in result:
+                log.write(ctx, 'export')
```
lwesterhof: Leftover debug statements

intake_vault.py
```diff
@@ -138,8 +138,7 @@ def dataset_collection_move_2_vault(ctx, toplevel_collection, dataset_id, vault_
     buffer = {}
     if status == 0:
         # stamp the vault dataset collection with additional metadata
-        date_created = datetime.now()
-        avu.set_on_coll(ctx, vault_path, "dataset_date_created", date_created.strftime('%Y-%m-%dT%H:%M:%S.%f%z'))
+        avu.set_on_coll(ctx, vault_path, "dataset_date_created", str(int(time.time())))
```
lwesterhof, line 141: Does this change the way dataset_date_created is stored? How is this handled for existing datasets?

lwesterhof (changes requested):

intake_vault.py
HarmdR: Good observation and question!
The timestamp must be stored as a number. This is the way it was originally and I introduced this fuckup myself.
I came across this error when I added the reporting part of this module to the python portal.
I erroroneously added a string formatted (Ymd etc) dataset creation timestamp.

Originally, in youth cohort ruleset, it was:

				# stamp the vault dataset collection with additional metadata
				msiGetIcatTime(*date, "unix");
				msiAddKeyVal(*kv, "dataset_date_created", *date);

As 'Youth' is not working with the new python refactored rules yet , as it was not released yet, no datasets have been added with an erroneous dataset_creation (gelukkig!)  :-)

intake_dataset.py
HarmdR: well spotted! Totally overlooked them

Corrected 👍

lwesterhof (approved):

PR #79, opened 2021-06-15
Statistics: fix bug YDA-4103

stsnel:
Size of some groups was counted wrong when one group name was prefix
of another one, e.g. data objects of group "research-jansen" were
included in size of group "research-jan"

If PR is accepted, please cherry-pick to release-1.6 and release-1.7 branches as well.


lwesterhof (approved):
LGTM

PR #83, opened 2021-06-21
Fix YDA-4115

stsnel:
Size of revisions wasn't counted in statistics job because of
double slash in collection search path (UUREVISIONCOLLECTION already
starts with a slash, so doesn't need to be prefixed with a slash).

Also fixed potential issue where two group names with a common prefix
(e.g. research-jan and research-jansen) could result in a situation
where the contents of the group with the long name will be counted
towards the statistics of the group with the short name.

### DIFF
--- a/resources.py
+++ b/resources.py
@@ -539,8 +539,8 @@ def rule_resource_store_monthly_storage_statistics(ctx):
                         tier_storage[the_tier] += int(row[0])
 
             # 3) Revision erea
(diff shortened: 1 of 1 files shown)


stsnel:
If PR is accepted, please cherry-pick to release-1.6 and release-1.7 as well

lwesterhof (approved):
LGTM

PR #102, opened 2021-07-26
Refactored tests to work with new auth flow

reavdhoef:
Adjusted the test for oidc login

Changed the step used to login as a user.

Testing was erratic for me; sometimes some worked and sometimes they failed. I think this is because of the issues mentioned around timing. Please verify if I didn't accidentally break all the tests.

### DIFF
--- a/tests/features/ui/ui_login_oidc.feature
+++ b/tests/features/ui/ui_login_oidc.feature
@@ -3,7 +3,7 @@ Feature: Login OIDC UI
 
     Scenario Outline: Login using OIDC
         Given user is not logged in
-        And login page is shown
+        And the user "<user>" can start the OIDC flow
         When user clicks login with OIDC
         And user "<user>" follows OIDC login process
         Then user "<user>" is logged in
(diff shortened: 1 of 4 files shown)


tests/step_defs/ui/test_ui_login_oidc.py
```diff
@@ -12,11 +12,16 @@
 )
 
 scenarios('../../features/ui/ui_login_oidc.feature')
+portal_url = "https://portal.yoda.test"
```
lwesterhof: Can be replaced with: `from conftest import portal_url`

tests/features/ui/ui_login_oidc.feature
```diff
@@ -3,7 +3,7 @@ Feature: Login OIDC UI
 
     Scenario Outline: Login using OIDC
         Given user is not logged in
-        And login page is shown
+        And the user "<user>" can start the OIDC flow
         When user clicks login with OIDC
```
lwesterhof, line 7: This step is not needed in the new flow?

lwesterhof (changes requested):

lwesterhof:
Not all UI tests work for me after these changes, ui_login_redirect tests fail.

PR #106, opened 2021-08-12
YDA-4132: ability to stop revision/replication jobs

lwesterhof:
Can be tested with:
```
imkdir /tempZone/yoda/flags
itouch /tempZone/yoda/flags/stop_replication
itouch /tempZone/yoda/flags/stop_revisions
```


stsnel (changes requested):
Thanks! Could you also add the flag check to the main data object loops of the revision and replication batch rules? This ensures that the rule will terminate gracefully after processing the current data object if the flag is set during execution of the rule.

stsnel (approved):
LGTM, thanks!

PR #107, opened 2021-08-13
YDA-4209 workaround for GenQuery issue with delete

stsnel:
This is a partial workaround for GenQuery bugs that result in
errors when called indirectly by the deletion policies
(see YDA-4209). By first checking that the user is a rodsUser
(not an admin), this will short circuit the queries for admins,
so that admins can circumvent GenQuery bugs when deleting
objects/collections in the research space.

If accepted, please cherry-pick into release-1.7 branch as well.


lwesterhof (approved):

lwesterhof:
LGTM

PR #115, opened 2021-09-29
YDA-4216:trigger policies on resources not in tree

stsnel:
If accepted, please also merge into release-1.7 branch
----

This adds a configuration option resource_trigger_pol to match
resources where data object modifications should trigger policies,
so that policies can be enabled on resources other than the primary
resource.

This change also fixes the situation that not configuring a replication
resource will have a side effect of disabling revision and metadata import
policies.


lwesterhof (approved):
LGTM!

PR #117, opened 2021-10-06
Fix import lint warnings for intake rules

stsnel:


stsnel:
This PR is no longer needed, has been fixed in the mean time.

PR #124, opened 2021-11-02
Pyportal stats more robust

HarmdR:
-Extended the logging of the data storage collection process
-When collecting storage (on group level) throws an exception then for that specific group the storage data of the previous month is taken

### DIFF
--- a/resources.py
+++ b/resources.py
@@ -5,6 +5,7 @@ __copyright__ = 'Copyright (c) 2018-2021, Utrecht University'
 __license__   = 'GPLv3, see LICENSE'
 
 from datetime import datetime
+from datetime import timedelta
 from math import ceil
 
 import genquery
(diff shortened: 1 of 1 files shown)


stsnel (changes requested):
Thanks, looks good to me! I have a few minor comments and questions:
- ~~Could you add a common prefix (e.g. "monthly_storage_statistics: " ) to the log messages, so that it is easier for admins to grep for log messages by this rule?~~
- As far as I'm concerned, the "Month current" and "Month previous" log messages can be removed. I don't think they're needed for application administration.
- I think we'll need to review and update the statistics module design docs at https://github.com/UtrechtUniversity/yoda/blob/development/docs/design/overview/statistics.md after this change. We could either do this as part of this ticket, or split it off to a different ticket, if that's more practical.

lwesterhof:
The common prefix should be handled by our log utility, write() will prefix the log string with originating rule/API name https://github.com/UtrechtUniversity/irods-ruleset-uu/blob/c0fe1956e49a67c63300e8f982074e56de9c8c2a/util/log.py#L15

So all messages should be prefixed with rule_resource_store_monthly_storage_statistics.

stsnel (approved):
LGTM!

PR #131, opened 2021-11-22
YDA-4395 intake scan: normalize experiment type

stsnel:
Normalize experiment type in names intake group to lowercase,
so that the intake-to-vault job does not create multiple
experiment type folders in the vault with different capitalizations.

If accepted, please also merge into release-1.7 branch.


HarmdR (approved):

lwesterhof (approved):
LGTM

PR #139, opened 2022-01-25
Development fix max char groupname

HarmdR:


group.py
```diff
@@ -418,8 +418,18 @@ def api_group_create(ctx, group_name, category, subcategory, description, data_c
...
     :param data_classification: Data classification of the group to create
+
+    :returns: Dict holding process status and process status info
     """
-    ctx.uuGroupAdd(group_name, category, subcategory, description, data_classification, '', '')
+    try:
+        response = ctx.uuGroupAdd(group_name, category, subcategory, description, data_classification, '', '')['arguments']
+        status = response[5]
+        message = response[6]
+        return {'proc_status': status,
```
lwesterhof: We should use the API responses instead of custom dicts, example: https://github.com/UtrechtUniversity/yoda-ruleset/blob/b72a51894131b096a95518e826769c013d24c8f7/research.py#L86

group.py
```diff
@@ -418,8 +418,18 @@ def api_group_create(ctx, group_name, category, subcategory, description, data_c
...
+                'proc_status_info': message}
+    except Exception:
+        return {'proc_status': 'NOK',
```
lwesterhof: We should use the API responses instead of custom dicts, example: https://github.com/UtrechtUniversity/yoda-ruleset/blob/b72a51894131b096a95518e826769c013d24c8f7/research.py#L84

group.py
```diff
@@ -430,8 +440,18 @@ def api_group_update(ctx, group_name, property_name, property_value):
...
     :param property_value: Value of the property to update
+
+    :returns: Dict holding process status and process status info
     """
-    ctx.uuGroupModify(group_name, property_name, property_value, '', '')
+    try:
+        response = ctx.uuGroupModify(group_name, property_name, property_value, '', '')['arguments']
+        status = response[3]
+        message = response[4]
+        return {'proc_status': status,
```
lwesterhof: We should use the API responses instead of custom dicts. Inside the `try` we can probably return 'api.Result.ok()' if status is 1. If the status is 0 or not 1 we can return the correct error message with api.Error. The exception returns all further unhandled errors.

lwesterhof (changes requested):

group.py
HarmdR: Gedaan en werkt idd heel mooi!
Het maakt de veranderingen in yoda-portal idd niet meer noodzakelijk.

Gr,
Harm

group.py
```diff
@@ -418,8 +418,19 @@ def api_group_create(ctx, group_name, category, subcategory, description, data_c
...
+    try:
+        response = ctx.uuGroupAdd(group_name, category, subcategory, description, data_classification, '', '')['arguments']
+        status = response[5]
+        message = response[6]
+        if status == '0':
+            return api.Result.ok()
+        else:
+            return api.Error('-1', message)
+    except Exception:
+        return api.Error('-1', 'Something went wrong creating group "{}". Please contact a system administrator'.format(group_name))
```
lwesterhof: Instead of -1 we can give the error a name to influence the http status returned by the portal (https://github.com/UtrechtUniversity/yoda-portal/blob/26b233cc613ba0227de482f4ac16c23ab25312a3/api.py#L30).

When an exception occurs, we can use the name 'error_internal'. This will result in http status 500.

group.py
```diff
@@ -418,8 +418,19 @@ def api_group_create(ctx, group_name, category, subcategory, description, data_c
...
     """
-    ctx.uuGroupAdd(group_name, category, subcategory, description, data_classification, '', '')
+    try:
+        response = ctx.uuGroupAdd(group_name, category, subcategory, description, data_classification, '', '')['arguments']
+        status = response[5]
+        message = response[6]
+        if status == '0':
+            return api.Result.ok()
+        else:
+            return api.Error('-1', message)
```
lwesterhof: Instead of -1 we can give the error a name to influence the http status returned by the portal (https://github.com/UtrechtUniversity/yoda-portal/blob/26b233cc613ba0227de482f4ac16c23ab25312a3/api.py#L30).

When an policy error occurs, we can use the name 'policy_error' (something other than error_internal and ok). This will result in http status 400. We want http status 400, because this was an invalid request from the client to the server.

lwesterhof (changes requested):

group.py
HarmdR: Changed status into 'error_internal' and 'policy_error' at the proper locations 👍

lwesterhof (approved):

PR #152, opened 2022-03-04
Increase lock time to 1 week

stsnel:
This is part of the fix for YDA-4407, which improves locking for the
intake-to-vault job. The main part of the fix is in the Ansible
playbook. We're increasing the lock time so that the lock remains
in place during typical batches, which could take more than 5 minutes
to process, but typically not more than a couple of hours.

If accepted, please merge into release-1.7 and release-1.8 branches as well.


lwesterhof (approved):
LGTM

PR #180, opened 2022-07-30
added license including uri

HarmdR:
added license/uri to rightslist.
The license can be found in the dataCite output (after publication).
But the rightsURI does not show up in this dataCite output - can't find the cause of this and am contacting datacite about this


lwesterhof (approved):
LGTM

PR #182, opened 2022-08-02
Intake group name format mismatch between intake / GM module.

HarmdR:
finally een streep eronder 👍


policies_intake.py
```diff
@@ -184,3 +202,10 @@ def coll_in_path_of_locked_dataset(ctx, actor, coll):
 
         # There is no lock present
         return False
+
+
+def _get_intake_group(coll):
+    """ Get the group type defining part of the entire groupname. 'grp-intake' or 'intake' """
+    parts = coll.split('/')[3].split('-')
```
stsnel, line 209: Two suggestions regarding this function:
- I think it would improve readability of other rules/functions that call this function if the name is a bit more specific, e.g. _get_intake_group_prefix or _get_intake_group_type. 
- Perhaps consider validating whether the input parameter has the expected format, and throwing a ValueError or similar if not. That would make it easier to troubleshoot any issues that involve other functions that pass incorrect values to this function.

intake.py
```diff
@@ -488,7 +503,11 @@ def api_intake_lock_dataset(ctx, path, dataset_ids):
     # check permissions - datamanager only
     parts = path.split('/')
     group = parts[3]
+
+    # grp-intake-FOO => grp-datamanager-FOO
     datamanager_group = group.replace("-intake-", "-datamanager-", 1)
```
stsnel, line 508: - The code fragment for converting an intake group name to a datamanager group name is duplicated a few times in this file. Could be a candidate for extracting a function.
- When extracting  a function, it would also be useful to do some input validation, e.g. something like
```
if group.startswith("grp-intake-"):
   return group.replace("-intake-", "-datamanager-", 1)
elif group.startswith("intake-"):
   return group.replace("intake-", "grp-datamanager-", 1)
else:
   raise ValueError("Unexpected intake group format for group "+group)
```
This would also take care of weird edge cases (e.g. "intake-intake-youth")

stsnel:
The code in this PR has been copied to #195. Therefore closing this PR.

PR #200, opened 2022-09-08
YDA-4824: add API call to get token lifetime

stsnel:
So that the portal can show configured token lifetime to users


stsnel:
We've agreed to fix the issue internally in yoda-portal, so that an API call is no longer required. Therefore closing this PR.

PR #202, opened 2022-09-23
Decouple notifications logic from mail logic.

lwesterhof:
Note that https://github.com/UtrechtUniversity/yoda-ruleset/blob/5bc13fd8daa8a24ea9295ed3e169ab96242d5bc3/mail.py#L32 is still there. The configuration item is called notifications_enabled, it indicates if smtp is configured or not. This is because before notifications in the portal was introduced the e-mails were called notifications.

### DIFF
--- a/mail.py
+++ b/mail.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 """Rules for sending e-mails."""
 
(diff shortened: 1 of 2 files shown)


stsnel (approved):
Great! LGTM

PR #207, opened 2022-10-27
Development grouplevel schemas

HarmdR:
Hi Lazlo,
Please take a look at this for adding groups getting the right schema-id.


groups.py
```diff
@@ -768,6 +790,25 @@ def api_group_subcategories(ctx, category):
     return getSubcategories(ctx, category)
 
 
+@api.make()
```
lwesterhof: This API call fits better in schema.py I think.

schema.py
```diff
@@ -28,6 +28,19 @@ def get_group_category(callback, rods_zone, group_name):
...
+    iter = genquery.row_iterator(
+        "META_USER_ATTR_VALUE",
+        "USER_NAME = '{}' AND USER_TYPE = 'rodsgroup' AND META_USER_ATTR_NAME = 'schema_id'".format(group_name),
+        genquery.AS_LIST, callback
+    )
+
+    for row in iter:
+        # return found schemaCategory directly on group level
+        # No further test is required here as the value found here was selected from /rodszone/yoda/schemas/ and therefore MUST be present
+        log.write(callback, "FOund GROUP level schema_id: " + row[0])
```
lwesterhof: Leftover debug statement

uuSudoPolicies.r
```diff
@@ -51,6 +51,7 @@ acPreSudoUserRemove(*userName, *policyKv) {
 
 acPreSudoGroupAdd(*groupName, *initialAttr, *initialValue, *initialUnit, *policyKv) {
 	writeLine("serverLog", "In acPreSudoGroupAdd, group is <*groupName>, actor is <$userNameClient#$rodsZoneClient>");
+
```
lwesterhof: Leftover whitespace

lwesterhof (changes requested):
The functions in schema.py, for example get_active_schema_path, should also be adjusted to first check for a schema_id on the research group. Otherwise the wrong schema will be loaded when opening the metadata form.

[HarmdR] 
No action required here.
I've tested it all up until and including the stage of publication 👍

Brief explanation:

get_active_schema_path relies on get_group_category - which is adjusted for group level schema-ids

get_active_schema relies on get_active_schema_path and that depends on get_group_category

Etc etc

groups.py
HarmdR: I guess you meant get_schemas here which indeed is moved from groups.py to schem.py 👍

schema.py
HarmdR: fixed

uuSudoPolicies.r
HarmdR: fixed

PR #212, opened 2022-11-10
Fix error message when checksumming failed

fiberbit:
A probable copy-paste error in the past. Quite confusing in the logs for an administrator.


lwesterhof:
Thanks!

PR #220, opened 2022-12-07
Updates to vollmer-0

lwesterhof:


stsnel (commented):
It appears that the tag placeholder ("select a tag ...") and the funding reference help text ("ask your supervisor ...") are not visible in the metadata form, even though they have been added to the schema.

Perhaps this information can be moved to the tag help text and funder name help text as a workaround? Should we also register findings for this behaviour of the metadata form?

Apart from this, it looks good to me.

lwesterhof:
Thanks, both should be fixed now. Tag placeholder was set on the array and not on the array items. And the funding reference help text should have been an object description instead of a field help text.

stsnel:
Thanks for the explanation, the new version looks good to me.

PR #232, opened 2022-12-22
Use _wrapper from mail.py

fj-morales:
The daily notifications fail as the _wrapper function is defined in mail.py


lwesterhof:
Is it possible to rebase this PR against the development branch? Then we can cherry-pick the fix to the release-1.8 branch after merging. Thanks!

lwesterhof:
Fixed in #238

PR #234, opened 2023-01-02
Development stats groups datatables

HarmdR:
api functie om datatables (conform research browsing) te voeden in stats module


HarmdR:
resources.py was changed and should be reviewed.
I applied the shorthand to the list to do the slicing/pagination as discussed earlier

PR #239, opened 2023-01-06
Add retry for replication of data objects on resources that are remov…

lwesterhof:
…ed from the resource hierarchy).

### DIFF
--- a/replication.py
+++ b/replication.py
@@ -57,7 +57,7 @@ def rule_replicate_batch(ctx, verbose):
 
         # Get list of data objects scheduled for replication.
         iter = genquery.row_iterator(
-            "ORDER(DATA_ID), COLL_NAME, DATA_NAME, META_DATA_ATTR_VALUE",
+            "ORDER(DATA_ID), COLL_NAME, DATA_NAME, META_DATA_ATTR_VALUE, DATA_RESC_NAME",
             "META_DATA_ATTR_NAME = '{}'".format(attr),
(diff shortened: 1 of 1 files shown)


stsnel (commented):
Perhaps we can improve the robustness of this rule by adding checks in the fallback logic to ensure that we are always replicating from the right source replica if the source resource listed in the replication AVU is out of date. I think adding a check that the resource location of the source replica is on the provider and that the source replica is the one with the latest modification date would cover all cases (including data objects with multiple data objects on the provider).

replication.py
```diff
@@ -100,10 +101,20 @@ def rule_replicate_batch(ctx, verbose):
...
             except msi.Error as e:
                 log.write(ctx, 'ERROR - The file could not be replicated: {}'.format(str(e)))
+
+                # Retry replication with data resource name (covers case where resource is removed from the resource hierarchy).
                 try:
-                    ctx.msi_add_avu('-d', path, errorattr, "{},{}".format(from_path, to_path), "")
-                except Exception:
-                    pass
+                    # Workaround the PREP deadlock issue: Restrict threads to 1.
+                    ofFlags = "numThreads=1++++rescName={}++++destRescName={}++++irodsAdmin=++++verifyChksum=".format(data_resc_name, to_path)
```
stsnel, line 112: From an application admin perspective, I think it would be useful to log that the fallback replication behaviour has been triggered, since this should be an unusual situation.

PR #243, opened 2023-01-27
YDA-4925: add API endpoint delete expired DAPs

stsnel:
Add API endpoint to delete expired data access passwords. Called from portal in order to prevent label conflicts between newly created tokens and expired tokens that are no longer visible to user.


lwesterhof (approved):
LGTM, can you add an API test for the new endpoint?

PR #246, opened 2023-01-31
Prevent renaming a file in the portal while it is being written

lwesterhof:
This prevents leaving the file in a stale state.

### DIFF
--- a/research.py
+++ b/research.py
@@ -471,10 +471,10 @@ def api_research_file_rename(ctx, new_file_name, coll, org_file_name):
     if data_object.exists(ctx, path_target):
         return api.Error('invalid_destination', 'The selected filename ' + new_file_name + ' already exists')
 
-    # Data object in good state? (replica is not actively being written to or opened for read)
-    if not data_object.has_replica_with_status(ctx, coll + '/' + org_file_name, [constants.replica_status.INTERMEDIATE_REPLICA,
-                                                                                 constants.replica_status.READ_LOCKED,
-                                                                                 constants.replica_status.WRITE_LOCKED]):
+    # Does data object has replica in intermediate or locked state? (replica is not actively being written to or opened for read)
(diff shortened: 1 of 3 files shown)


research.py
```diff
@@ -471,6 +471,10 @@ def api_research_file_rename(ctx, new_file_name, coll, org_file_name):
     if data_object.exists(ctx, path_target):
         return api.Error('invalid_destination', 'The selected filename ' + new_file_name + ' already exists')
 
+    # Data object in good state? (replica is not actively being written to or opened for read)
+    if not data_object.replica_status(ctx, coll + '/' + org_file_name) == constants.replica_status.GOOD_REPLICA:
```
stsnel: This check could fail if the data object has a replica in a stale state, even though that should not cause any issues when renaming it.

util/data_object.py
```diff
@@ -42,6 +42,18 @@ def size(ctx, path):
...
 
+def replica_status(ctx, path):
+    """Get a data object's replica status."""
+    iter = genquery.row_iterator(
+        "DATA_REPL_STATUS",
+        "COLL_NAME = '%s' AND DATA_NAME = '%s'" % pathutil.chop(path),
+        genquery.AS_LIST, ctx
+    )
+
+    for row in iter:
```
stsnel, line 60: This function would always return the replica status of the first replica returned by the GenQuery. I think the function could be written more defensively by checking all replicas. Something like:

```python
def dataobject_has_replica_with_status(ctx, path, statuses):
  # ...
  for row in iter:
    if int(row[0]) in statuses:
        return True
  
  return False
```

Then this function could be called to check if the data object has any intermediate or locked replicas.

stsnel (commented):
It looks good to me in general; have added two comments regarding possible edge cases.

stsnel (approved):
LGTM!

PR #247, opened 2023-02-02
intake: handle unknown creation info

stsnel:
In the intake module, make intake_dataset_get_details return "unknown" if no creation information is available, rather than failing with KeyError.


lwesterhof (approved):
LGTM

PR #249, opened 2023-02-06
YDA-5048: fix silent failures intake module

stsnel:
Fix silent failures on intake module API calls for locking, unlocking and adding a comment to a dataset. Also add API tests for this scenario.


lwesterhof (approved):
Screenshot_2023-02-07_14-51-31

LGTM

PR #250, opened 2023-02-06
YDA-5036: Fix error handling secure copy

stsnel:
Update the publication logic so that it takes into account error codes returned by the msiExecCmd microservice used for transferring data to the public server. The previous logic always set the error code to 0, because of the interaction between the legacy rule language secure-copy rule and the rest of the logic in the PREP.


lwesterhof (approved):
LGTM!

PR #256, opened 2023-02-22
YDA-5076: do not create revisions for data objects moved outside rese…

lwesterhof:
…arch space.

### DIFF
--- a/revisions.py
+++ b/revisions.py
@@ -352,7 +352,7 @@ def rule_revision_batch(ctx, verbose):
                 avu.set_on_data(ctx, path, errorattr, "true")
 
         # Total revision process completed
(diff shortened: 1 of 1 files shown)


stsnel (approved):
Looks good! 

I noticed that the revision job output in the rodsLog says that `0 objects [were] ignored` if deleted data objects marked for revision were present. Code fragment: https://github.com/UtrechtUniversity/yoda-ruleset/blob/d39dca408aaa375d8e56eb2da864a159d295a6dc/revisions.py#L355

Perhaps we can make clearer that deleted data objects do not count towards ignored objects in this context. Example message: `0 data objects in research area ignored`.

PR #257, opened 2023-02-28
Development gm group list

HarmdR:
Added new setting to be able to save preferred view (tree or flat list) in group manager


lwesterhof (approved):
LGTM

PR #258, opened 2023-03-09
Development gm test flat list

HarmdR:
This test requires helper function in group_manager.js 
See my mail as well 👍


lwesterhof (approved):
LGTM

Screenshot_2023-03-10_11-40-27

PR #259, opened 2023-03-09
per category ordered on subcategory

HarmdR:
This was the shortest solution I could come up with
Possibly there's a smarter way


lwesterhof (approved):
LGTM

PR #261, opened 2023-03-22
YDA-4477 - Modified the script for monthly storage statistics to include and export all the categories.

kaur16:
YDA-4477 - Modified the script for monthly storage statistics to include and export all the categories.


lwesterhof (approved):
LGTM

PR #262, opened 2023-03-28
Persist publication state after errors in checking DOI availability

fiberbit:
This makes the "retry" process working better, as otherwise the publication can get into an endless loop.

This bug could happen when posting a DOI did work, but the acknowledgement run into a timeout and wasn't received. The fact that we did already try an upload wasn't registered then in the status. On the next retry we then hit an unexpected status in the state machine and again went to "retry".


stsnel (approved):
LGTM, thanks for the fix!

PR #265, opened 2023-03-31
YDA-5132 - Modify the groups section on the Statistics page to include intake, grp and datarequest groups.

kaur16:
YDA-5132 - Modify the groups section on the Statistics page to include intake, grp and datarequest groups.


lwesterhof (approved):
LGTM!

PR #266, opened 2023-03-31
YDA-5149: improve subcategory performance

stsnel:
Optimize genquery for finding category/subcategory to use default iCAT database indexes. This improves subcategory autocompletion performance in the group manager substantially on large environments.


lwesterhof (approved):
LGTM

PR #268, opened 2023-04-04
Add datarequest AVU syncer

aucuparia:


lwesterhof (changes requested):
This can be triggered by a policy, no need to manually trigger this synchronization: https://github.com/UtrechtUniversity/yoda-ruleset/blob/73dce3230839a5907bfd4392659385dc75c90321/policies.py#L517-L532

Adding a condition for the datarequest metadata and calling datarequest_sync_avus would be sufficient.

aucuparia:
> This can be triggered by a policy, no need to manually trigger this synchronization:
> 
> https://github.com/UtrechtUniversity/yoda-ruleset/blob/73dce3230839a5907bfd4392659385dc75c90321/policies.py#L517-L532
> 
> Adding a condition for the datarequest metadata and calling datarequest_sync_avus would be sufficient.

Good one, that's much more convenient. I've removed the wrapper script in tools and added a policy.

datarequest.py
```diff
@@ -700,6 +701,42 @@ def rule_datarequest_review_period_expiration_check(ctx):
         datarequest_process_expired_review_periods(ctx, [result['COLL_NAME'].split('/')[-1] for result in list(qcoll)])
 
 
+@rule.make(inputs=[0], outputs=[1])
```
lwesterhof: We can remove the rule decorator, because we don't have to call this from the rule language.

policies.py
```diff
@@ -530,6 +530,11 @@ def pep_resource_modified_post(ctx, instance_name, _ctx, out):
             # Path is a metadata file, ingest.
             log.write(ctx, 'metadata JSON <{}> modified by {}, ingesting'.format(path, username))
             ctx.rule_meta_modified_post(path, username, zone)
+        elif (info.space is pathutil.Space.DATAREQUEST
+              and pathutil.basename(info.subpath) == datarequest.DATAREQUEST + datarequest.JSON_EXT):
+            request_id = pathutil.dirname(info.subpath)
+            log.write(ctx, 'datarequest JSON <{}> modified by {}, ingesting'.format(path, username))
+            ctx.rule_datarequest_sync_avus(request_id)
```
lwesterhof: We can directly call the Python rule here.

lwesterhof (approved):
LGTM

stsnel (approved):
LGTM!

PR #274, opened 2023-04-17
Optimize query in get_group_category_info

stsnel:
Using IN instead of wildcard enables use of default iCAT database indexes, which speeds up the query on large environments.


lwesterhof (approved):
LGTM, we could make the same optimization in uuGroupGetCategory (uuGroup.r). That could speedup uuGroupModify.

PR #275, opened 2023-04-17
YDA-5023 - Set Yoda flags to use Collections on the Yoda flags system collection so that replication and revision jobs are not triggered

kaur16:
YDA-5023 - Set Yoda flags to use Collections on the Yoda flags system collection so that replication and revision jobs are not triggered


stsnel (approved):
Great, looks good 👍

PR #280, opened 2023-05-10
DRM schema versioning

aucuparia:


datarequest.py
```diff
@@ -554,22 +556,30 @@ def datarequest_reviewers_get(ctx, request_id, pending=False):
...
+def datarequest_schema_get(ctx, schema_name, version = SCHEMA_VERSION):
     """Get schema and UI schema of a datarequest form
 
     :param ctx:         Combined type of a callback and rei struct
     :param schema_name: Name of schema
+    :param version:     Version of schema
 
     :returns: Dict with schema and UI schema
     """
+    log.write(ctx, '=======================================')
```
lwesterhof: Debug statements can be removed?

datarequest.py
```diff
@@ -66,7 +66,9 @@
 
 JSON_EXT          = ".json"
 
-SCHEMACOLLECTION  = constants.UUSYSTEMCOLLECTION + "/datarequest/schemas/youth-0"
+SCHEMACOLLECTION  = constants.UUSYSTEMCOLLECTION + "/datarequest/schemas"
+SCHEMA_URI_PREFIX = "https://yoda.uu.nl/datarequest/schemas/"
+SCHEMA_VERSION    = "youth-1"
```
lwesterhof, line 71: In a later iteration we can make this a configuration option, so that the schema is not hard coded.

lwesterhof (changes requested):

datarequest.py
aucuparia: Added to backlog.

datarequest.py
aucuparia: Done.

aucuparia (commented):
All issues solved.

lwesterhof (approved):
Because one of the schemas is a symlink we get an error, because the two schemas have the same name.

Don't the other schemas like assignment and datamanager review need a schema id as well?

aucuparia:
> Because one of the schemas is a symlink we get an error, because the two schemas have the same name.
> 
> Don't the other schemas like assignment and datamanager review need a schema id as well?

Symlink is gone now, and I've added $id fields to the other schemas as well.

lwesterhof (approved):
LGTM

PR #281, opened 2023-05-10
YDA-5130: support for parallel replication/revision jobs

lwesterhof:
Ansible changes in https://github.com/UtrechtUniversity/yoda/pull/258


replication.py
```diff
@@ -53,26 +63,35 @@ def rule_replicate_batch(ctx, verbose):
...
-            "META_DATA_ATTR_NAME = '{}'".format(attr),
-            genquery.AS_LIST, ctx
-        )
+        log.write(ctx, "Batch replication job started - balance id: {}-{}".format(balance_id_min, balance_id_max))
+
+        # Get list up to 1000 data objects scheduled for replication.
+        iter = list(genquery.Query(ctx,
+                    ['ORDER(DATA_ID)', 'COLL_NAME', 'DATA_NAME', 'META_DATA_ATTR_VALUE', 'DATA_RESC_NAME'],
+                    "META_DATA_ATTR_NAME = '{}'".format(attr),
+                    offset=0, limit=1000, output=genquery.AS_LIST))
```
stsnel: We might want to adjust the batch sizes of the replication and revision job based on experiences with using the new implementation in production. Can we extract configuration parameters for the limit values of the replication and revision job?

tools/async-job.py
```diff
@@ -13,47 +13,33 @@
...
+    rule_name = 'uuReplicateBatch(*verbose, *balance_id_min, *balance_id_max)'
 elif 'revision' in NAME:
-    rule_name = 'uuRevisionBatch(*verbose)'
+    rule_name = 'uuRevisionBatch(*verbose, *balance_id_min, *balance_id_max)'
 else:
     print('bad command "{}"'.format(NAME), file=sys.stderr)
     exit(1)
 
 args = get_args()
-lock_or_die()
```
stsnel, line 56: I believe we need to insert a replacement for the old locking function here, in order to reduce the risk of multiple revision/replication jobs running simultaneously with overlapping balance ID ranges.

replication.py
```diff
@@ -53,26 +63,35 @@ def rule_replicate_batch(ctx, verbose):
...
+            to_path = info[1]
+
+            # Backwards compatability with replication metadata created in v1.8 or earlier.
+            backwards_compatibility = False
+
+            if len(info) == 3:
+                balance_id = int(info[2])
+            elif len(info) == 2:
+                backwards_compatibility = True
+                balance_id = int(balance_id_min)
```
stsnel: Good that we have fallback logic for backwards compatibility here 👍  . I would suggest adjusting the fallback logic so that it deterministically assigns data objects with an old-format revision/replication AVU to a single revision/replication job, to prevent data objects from being processed by multiple jobs concurrently.

For example:

```
balance_id = int(hashlib.md5(path.encode('utf-8')).hexdigest(), 16) % 64 + 1
```

Same thing for the revision logic.

stsnel:
Looks good in general :+1: I have added a couple of comments about specific aspects.

I haven't tested the implementation yet at this time.

replication.py
HarmdR: good thinking 👍
This is implemented

tools/async-job.py
HarmdR: This is now implemented where the locking file is extended with the corresponding min/max balance_id. This, as this is directly deterministic for the job.

replication.py
HarmdR: This is implemented exaclty as suggested 👍

revisions.py
```diff
@@ -298,12 +306,36 @@ def rule_revision_batch(ctx, verbose):
             # Perform scheduled revision creation for one data object.
             data_id = row[0]
             path    = row[1] + "/" + row[2]
-            resc    = row[3]
+
+            test = int(hashlib.md5(path.encode('utf-8')).hexdigest(), 16) % 64 + 1
+            log.write(ctx, 'test hashlib: {}'.format(test))
```
stsnel: This test code could be removed

revisions.py
```diff
@@ -298,12 +306,36 @@ def rule_revision_batch(ctx, verbose):
...
+                resc = row[3]
+                # Determine a balance_id for this dataobject based on its path.
+                # This will determine whether this dataobject will be taken into account in this job/range or another that is running parallel
+                balance_id = int(hashlib.md5(path.encode('utf-8')).hexdigest(), 16) % 64 + 1
+
+            # Check whether balance id is within the range for this job.
+            if balance_id < int(balance_id_min) or balance_id > int(balance_id_max):
+                # Skip this one and go to the next data object for revision creation.
+                if print_verbose:
+                    log.write(ctx, "Batch revision: skipping {} due to load balancing: range {} {}".format(path, balance_id_min, balance_id_max))
```
stsnel: I think this log message can be removed now that we have tested the load balancing.

replication.py
```diff
@@ -83,8 +106,16 @@ def rule_replicate_batch(ctx, verbose):
                 # Go to next record and skip further processing.
                 continue
 
-            from_path = xs[0]
-            to_path = xs[1]
+            # Check whether balance id is within the range for this job
+            if balance_id < int(balance_id_min) or balance_id > int(balance_id_max):
+                # Skip this one and go to the next data object to be replicated.
+                if print_verbose:
+                    log.write(ctx, "Batch replication: skipping {} due to load balancing: range {} {}".format(path, balance_id_min, balance_id_max))
```
stsnel: I think this log message can be removed now that we have tested the load balancing.

stsnel (approved):
Great, looks good and works for me on dev  :+1: . Have only added a few small comments regarding log messages (see above)

PR #284, opened 2023-05-16
initial setup for renaming of monthly storage statistics job to stora…

HarmdR:
goes together with ansible branch with same name


lwesterhof (approved):
LGTM

PR #285, opened 2023-05-17
YDA-5208: fix missing language field landing page

stsnel:
Definition name for languages did not match schema.


lwesterhof (approved):
LGTM

PR #287, opened 2023-05-17
YDA-2547 - Get a user list with all the users that are not part of any group

kaur16:
YDA-2547 - Get a user list with all the users that are not part of any group


lwesterhof (approved):
LGTM

PR #290, opened 2023-05-31
YDA-5228: improve error handling vault transitions

stsnel:
If we encounter a folder that does not exist in the vault transition metadata, log a warning message in the serverLog and skip it, rather than crashing.


lwesterhof (approved):
LGTM

PR #291, opened 2023-06-01
Datacite: add TLS verification option

stsnel:
Need to be able to disable TLS verification for Datacite on development environments in order to communicate with the mock service (YDA-5117).


lwesterhof (approved):
LGTM

PR #294, opened 2023-06-09
fix due to incorrect determination of vault when subfolders are involved

HarmdR:
initially I thought it was caused by the javascript, but digging deeper showed that the determination of the vault path was wrong. As this path was not in existance, the vaultname remained empty causing the frontend to do 'strange things' :-)


lwesterhof (approved):
LGTM

PR #307, opened 2023-07-27
YDA-5037 - Changes to the depublication and republication process

kaur16:
YDA-5037 - Changes to the publication, depublication and republication process.


publication.py
```diff
@@ -229,7 +224,7 @@ def set_update_publication_state(ctx, vault_package):
     # Set publication status
     publication_state["status"] = "Unknown"
 
-    # Generate new JSONs
+    # Generate new XMLs
```
lwesterhof: Comment should be "Generate new JSONs".

publication.py
```diff
@@ -261,6 +256,8 @@ def get_publication_date(ctx, vault_package):
 
     :return: Publication date in ISO8601 format
     """
+    from datetime import datetime
```
lwesterhof: This import should be moved to the top-level imports.

publication.py
```diff
@@ -293,12 +289,12 @@ def get_last_modified_datetime(ctx, vault_package):
         "COLL_NAME = '" + vault_package + "' AND META_COLL_ATTR_NAME = '" + constants.UUORGMETADATAPREFIX + 'action_log' + "'",
         genquery.AS_LIST, ctx
     )
-
     for row in iter:
         log_item_list = jsonutil.parse(row[1])
-        my_date = datetime.fromtimestamp(int(log_item_list[0]))
 
-        # ISO8601-fy
+        import datetime
```
lwesterhof: With `from datetime import datetime` as top level import we can remove this import and change `datetime.datetime.fromtimestamp` below into `datetime.fromtimestamp`.

lwesterhof (changes requested):

PR #311, opened 2023-08-07
YDA-5319: implement delay times for async jobs

stsnel:
Implement optional delays after last modification before a data object is processed by the asynchronous replication and revision job. Setting a delay reduces the risk of consistency issues due to concurrent access bugs.


replication.py
```diff
@@ -67,10 +68,15 @@ def rule_replicate_batch(ctx, verbose, balance_id_min, balance_id_max, batch_siz
...
+        # Modification timestamps are recorded as varchar in iCAT, and GenQuery doesn't support casts, so
+        # we need to fill the minimum timestamp with leading zeroes in order to be able to compare against
+        # it in the following query.
+        minimum_timestamp = str(int(time.time()) - int(config.async_replication_delay_time)).zfill(11)
+
+        # Get list of up to 1000 data objects scheduled for replication, taking into account their modification time.
         iter = list(genquery.Query(ctx,
                     ['ORDER(DATA_ID)', 'COLL_NAME', 'DATA_NAME', 'META_DATA_ATTR_VALUE', 'DATA_RESC_NAME'],
-                    "META_DATA_ATTR_NAME = '{}'".format(attr),
+                    "META_DATA_ATTR_NAME = '{}' AND DATA_MODIFY_TIME <= '{}'".format(attr, minimum_timestamp),
```
lwesterhof, line 79: I don't think this is documented, but I think we can use `n<=` for comparison of numbers. Then we don't need to cast to string and fill with leading zeros.

PR #319, opened 2023-08-23
YDA-5309 - Added transformation script for existing publications.

kaur16:
YDA-5309 - Added transformation script for existing publications.


lwesterhof (approved):
LGTM

PR #320, opened 2023-08-23
YDA-5363 - Removed hardcoded code (previously used for backward compatibility) in the publication process

kaur16:
YDA-5363 - Removed hardcoded code (previously used for backward compatibility) in the publication process


lwesterhof (approved):
LGTM
Screenshot_2023-08-23_15-35-59

PR #321, opened 2023-08-23
Merge changes for release 1.8.9

stsnel:
Implement optional delays after last modification before a data object is processed by the asynchronous replication and revision job. Setting a delay reduces the risk of consistency issues due to concurrent access bugs.


lwesterhof (changes requested):
I think we can also add this commit: https://github.com/UtrechtUniversity/yoda-ruleset/commit/4d31b7d68a58db4aabaaae6291f95c480e4ddee2

PR #322, opened 2023-08-30
Development fix 5366

HarmdR:
-Proper initialization of variables (error only occurs when no storage data was registered yet)
-Intercept no data situation and return empty values directly. The frontend can properly react to that


stsnel (approved):
LGTM 👍

PR #323, opened 2023-08-30
Send notification to datamanager if research group has been inactive for many months

claravox:
Yda 2546

https://github.com/UtrechtUniversity/yoda/pull/298 depends on this pr


util/config.py
```diff
@@ -109,6 +109,8 @@ def __repr__(self):
                 enable_data_package_reference=False,
                 enable_tokens=False,
                 enable_tape_archive=False,
+                inactivity_cutoff_months=3,
+                inactivity_notification=0,
```
stsnel: Can you change the format of this configuration item so that it uses the same format as the other feature switch parameters, for consistency? Something like: enable_inactivity_notifications=False

These parameters can also be added to the `rules_uu.cfg.template` file, as well as The Docker setup ruleset config file (`docker/images/yoda_irods_icat/rules_uu.cfg`. )

notifications.py
```diff
@@ -344,6 +346,86 @@ def rule_process_groups_expiration_date(ctx):
...
+    iter = genquery.row_iterator(
+        "USER_GROUP_NAME",
+        "USER_TYPE = 'rodsgroup' AND USER_GROUP_NAME like 'research-%'",
+        genquery.AS_LIST, ctx
+    )
+
+    for row in iter:
+        group_name = row[0]
+        coll = '/{}/home/{}'.format(zone, group_name)
+        coll_general = '/{}/home/{}%'.format(zone, group_name)
```
stsnel: This wildcard matches subcollections (which is good) but also collections with the same prefix. For example, if you search for data objects in `/tempZone/home/research-foo%`, you will also get data objects in `/tempZone/home/research-foot`. Here's an example of code that matches with subcollections, but not other collections with the same prefix: https://github.com/UtrechtUniversity/yoda-ruleset/blob/a781dd81240236d851652b8b01d6ef94ee53b116/resources.py#L564-L579 .

notifications.py
```diff
@@ -344,6 +346,86 @@ def rule_process_groups_expiration_date(ctx):
...
+
+        # Second query: check if there are any files in the group, including files in the sub-folders
+        # If there are no files we do not want to notify about lack of activity
+        iter_data = genquery.row_iterator(
+            "DATA_NAME, COLL_NAME",
+            "COLL_NAME like '{}' AND USER_GROUP_NAME = '{}'".format(coll_general, group_name),
+            genquery.AS_LIST, ctx
+        )
+
+        if iter_data.total_rows() > 0:
```
stsnel: I think it would be best to add an else condition here that checks the `COLL_MODIFY_TIME` of the main collection if no data objects are currently present, so that we also send notifications for inactive groups that have no data objects anymore.

stsnel (changes requested):
Looks good in general 👍 I think the log messages will be especially useful for operations.

Have added a few comments about some specific parts.

notifications.py
claravox: One of the reasons for adding this notification feature was to encourage research groups to use less space. Should we still send the notification if no data objects are currently present, since in that case the group was not taking up a lot of space?

stsnel: That's a good point. I double checked with the stakeholder / data manager. He would also like to receive a notification if the group has been empty for three months. 

In that case it's more about reducing clutter in the environment, rather than clearing up storage space.

stsnel (approved):
Excellent, LGTM 👍

PR #324, opened 2023-08-31
removed hardcoded zone

HarmdR:
'tempZone' was used hardcoded as a zone reference


stsnel (approved):
Good catch 👍 

I would only suggest amending the commit message to include the ticket number (`YDA-5365`) in the commit, so that it is easier to see the context after merging.

PR #325, opened 2023-08-31
YDA-5379: tighten filter for deposit index coll

stsnel:
The portal should not display internal index collections in vault deposit groups.

This change improves the matching of collection names, so that collections named 'index' in archived data (e.g.
/tempZone/home/vault-deposit/deposit[123][456]/original/index) are shown normally , whereas internal index collections (e.g. /tempZone/home/vault-deposit/deposit[123][456]/index) are filtered out.


dworkin:
LGTM

dworkin (approved):

PR #332, opened 2023-09-26
extended possibility to go deeper into the datamanagergroup removal c…

HarmdR:
I think this fix solves YDA5392.
But I'm not 100% sure whether it intervenes with other situation


lwesterhof (approved):
LGTM

PR #340, opened 2023-10-05
YDA-4101: add ability to create external user with custom creator

claravox:
Requires https://github.com/UtrechtUniversity/yoda-clienttools/pull/5

### DIFF
--- a/uuGroup.r
+++ b/uuGroup.r
@@ -954,10 +954,11 @@ uuGroupUserAdd(*groupName, *user, *creatorUser, *creatorZone, *status, *message)
 	*status  = '1';
 	*message = "An internal error occurred.";
 
+	*fullNameActor = "$userNameClient#$rodsZoneClient";
 	# Check that the creator user exists
-	*fullName = "*creatorUser#*creatorZone";
+	*fullNameCreator = "*creatorUser#*creatorZone";
 
-	uuUserExists(*fullName, *exists);
+	uuUserExists(*fullNameCreator, *exists);
 	# If creator does not exist, exit
 	if (!*exists) {
     	succeed; # Return here (fail would ruin the status and error message).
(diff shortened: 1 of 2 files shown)


uuGroup.r
```diff
@@ -460,6 +460,25 @@ uuGroupGetDescription(*groupName, *description) {
...
+uuGroupGetMembers(*groupName, *members) {
+	uuGroupGetMembers(*groupName, false, false, *m);
+	*members = list();
+	foreach (*member in *m) {
+		# Throw away the zone name for backward compat.
+		uuChop(*member, *name, *_, "#", true);
+		*members = cons(*name, *members);
+	}
+}
+
```
claravox, line 481: Added back this function so that --delete and --allow-update options in yoda-clienttools would still work. Otherwise we do not use this function.

uuGroup.r
```diff
@@ -956,13 +986,13 @@ uuGroupUserAdd(*groupName, *user, *status, *message) {
...
                         *message = ""
-                        rule_group_provision_external_user(*userName, $userNameClient, $rodsZoneClient, *http_code, *message);
+                        rule_group_provision_external_user(*userName, *creatorUser, *creatorZone, *http_code, *message);
                         if (*message != "") {
                                 writeLine("serverLog", "[EXTERNAL USER] *message");
                                 *status = *http_code;
                                 succeed; # Return here (fail would ruin the status and error message).
                         }
-                        writeLine("serverLog", "[EXTERNAL USER] User *userName added by $userNameClient on $rodsZoneClient.");
+                        writeLine("serverLog", "[EXTERNAL USER] User *userName added on the behalf of *creatorUser on *creatorZone.");
```
stsnel: Good idea to log the creatorUser and creatorZone here. Can you also include the $userNameClient / $rodsZoneClient in the log message, so that admins can look up in the logs who originally created a user (even if it was on behalf of someone else)?

stsnel (changes requested):
It looks good in general :+1:  can you add an additional authorization check in case someone adds a group member on behalf of someone else? This function should only be available to rodsadmins, rather than any group admin.

stsnel (approved):
Great, LGTM :+1: 

Can you squash the commits to the Github Workflow before merging?

PR #341, opened 2023-10-09
YDA-5470 - Tested invitation flows in SRAM acceptance environment

kaur16:
YDA-5470 - Tested invitation flows in SRAM acceptance environment


lwesterhof (approved):
LGTM

PR #343, opened 2023-10-10
YDA-5414 - Added function to generate unique short name for groups.

kaur16:
YDA-5414 - Added function to generate unique short name for groups.


lwesterhof (approved):
LGTM

PR #347, opened 2023-10-16
YDA-5345: implement ARB

stsnel:
Automatic resource balancing ensures that new data objects are only created on resources that have enough space available.

See /docs/design/processes/automatic-resource-balancing.md in the Yoda documentation for design information.


resources.py
```diff
@@ -641,6 +643,78 @@ def rule_resource_store_storage_statistics(ctx):
...
+
+    if not resource.exists(ctx, resc_name):
+        log.write(ctx, "Error: could not find resource named '{}' for ARB update.".format(resc_name))
+        return
+
+    bytes_free_gb = int(bytes_free) / 2 ** 30
+    bytes_free_percent = 100 * (float(bytes_free) / float(bytes_total))
+
+    if resc_name in config.arb_exempt_resources:
+        arb_status = "EXEMPT"
```
lwesterhof: Can we define all the possible ARB status codes as enum in constants.py?

lwesterhof (approved):
LGTM!

resources.py
stsnel: Good idea, have changed it.

PR #362, opened 2023-10-26
YDA-5415: optimize revision cleanup

stsnel:
This optimizes the revision cleanup job by applying a pre-filter to collected revisions in the scan phase. If the number of revisions of a data object does not exceed the minimum bucket size, we already know these revisions won't be cleaned up, and we don't need to process them further.

Logic and strategy functions have been moved to a separate file, in order to facilitate adding unit tests.


lwesterhof (approved):
Nice, LGTM!

PR #366, opened 2023-11-03
YDA-5516: clean up old pre-bucket revisions

stsnel:
In the revision cleanup job, remove all revisions before the defined buckets, but ensure the last revision of a versioned data object is not removed, even if it is not in a defined bucket.

This ensures that a user can always restore a data object to an earlier version, but that we don't keep (many) more revisions than specified by the revision strategy.


lwesterhof (approved):
LGTM!

PR #371, opened 2023-11-27
YDA-5533: Revisions, replications specify max memory, dry run option

claravox:
Requires https://github.com/UtrechtUniversity/yoda/pull/325


replication.py
```diff
@@ -195,6 +221,32 @@ def is_replication_blocked_by_admin(ctx):
 
     :returns: Boolean indicating if admin put replication on hold.
     """
-    zone = user.zone(ctx)
-    path = "/{}/yoda/flags/stop_replication".format(zone)
-    return collection.exists(ctx, path)
+    return data_object.exists(ctx, "/{}{}".format(user.zone(ctx), "/yoda/flags/stop_replication"))
```
fiberbit: This reverts an earlier change (for Yoda 1.9) in the way the check if we should stop is done. It changed checking a data-object to checking a collection. Unfortunately using the exact name.

revisions.py
```diff
@@ -410,11 +434,9 @@ def is_revision_blocked_by_admin(ctx):
 
     :param ctx: Combined type of a callback and rei struct
 
-    :returns: Boolean indicating if admin put replication on hold.
+    :returns: Boolean indicating if admin put revisions on hold.
     """
-    zone = user.zone(ctx)
-    path = "/{}/yoda/flags/stop_revisions".format(zone)
-    return collection.exists(ctx, path)
+    return data_object.exists(ctx, "/{}{}".format(user.zone(ctx), "/yoda/flags/stop_revisions"))
```
fiberbit: This reverts an earlier change (for Yoda 1.9) in the way the check if we should stop is done. It changed checking a data-object to checking a collection. Unfortunately using the exact name.

util/config.py
fiberbit, line 1: Two new config values are added, but it might be a good idea to also add a value for "batch_size_limit"? The default of which is now only indirectly specified in "async-job.py".

fiberbit (commented):
See notes about changed behaviour in checking for "stop" data-objects or collections.

Note that a lot of things have changed in the Yoda-1.9 lifecycle so far, and none of them have been tested on production scale. Some "fixes" for issues are perhaps even no longer necessary or even wanted due to later changes. But that is a far more general issue and is not directly relevant to this patch set.

util/config.py
claravox: We have config items `async_revision_batch_size` and `async_replication_batch_size`. They are used in the [configuration](https://utrechtuniversity.github.io/yoda/administration/configuring-yoda.html#replication-configuration) of the cron jobs for replication and revision.

lwesterhof (approved):
LGTM

PR #385, opened 2024-01-04
YDA-5561 - Excluded 'expiration date' from SRAM API

kaur16:
YDA-5561 - Excluded 'expiration date' from SRAM API


lwesterhof (approved):
LGTM

PR #393, opened 2024-02-13
Yda 5575 fileviewer

claravox:


vault.py
```diff
@@ -607,6 +609,7 @@ def api_vault_collection_details(ctx, path):
         "basename": basename,
         "status": status,
         "metadata": metadata,
+        "member_type": member_type,
```
claravox, line 619: Do we have api versions? Would adding this break anything?

browse.py
```diff
@@ -315,3 +317,46 @@ def _filter_vault_deposit_index(row):
...
+
+    # Check end extension
+    valid_extension = False
+    for ext in config.text_file_extensions:
+        ending = '.' + ext
+        if file_path.lower().endswith(ending):
+            valid_extension = True
+
+    if not valid_extension:
+        return api.Error('not_valid', 'The given data object does not have a valid file extension')
```
claravox, line 349: I'm not sure how to pick names for the api errors.

lwesterhof: I think you did great :)

vault.py
lwesterhof: API version is in sync with the release version and can be found here: https://github.com/UtrechtUniversity/yoda-ruleset/blob/development/__init__.py#L4

Could be, but I don't think it does in this case.

lwesterhof:
Screenshot_2024-02-14_08-47-18

lwesterhof (approved):
LGTM!

PR #361, opened 2023-10-26
Policies: prevent changing or removing replication or revision metadata

lwesterhof:

PR #394, opened 2024-02-16
Extract Python 3 interpreter path parameter

stsnel:
So that we can use the ruleset on distributions that have python3 in different places (initially used for adding Ubuntu 20.04 LTS support)

(depends on https://github.com/UtrechtUniversity/yoda/pull/361)


lwesterhof (approved):
LGTM

PR #396, opened 2024-02-19
YDA-5605: make Scopus ID validation less strict

### LINKED ISSUE #362: [BUG] Yoda metadata default-3 incorrect scopus validation
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Within metadata default-3, within the 'creator' type only, the "Author identifier (Scopus)" requires a "name_identifier" of 11 numbers when there are identifiers available with just 10 numbers (for example https://www.scopus.com/authid/detail.uri?authorId=8375574300). Anything of lower than 11 is not allowed as an answer field and will not be saved. When converting from default 2 to default 3 within the webportal, an error will be thrown that conversion is also not possible. 
For contributor types, the number length is not enforced and is allowed to be saved when numbers are less than 11.

### Expected Behavior

Allow Author ID to be 10 numbers for creator types.

### Steps To Reproduce

Open Yoda metadata default 3.
Add creator with "Name_Identifier_Scheme": "Author identifier (Scopus)".
Try to add 10 numbers and save.
(or create a default 2 scheme, add creator with author ID scopus and add 10 numbers, then convert to default 3)

### Environment

```markdown
- Yoda:1.9 rc5 and rc6
```

stsnel:
Scopus Helpdesk confirmed that there is no standard for the number of digits in a Scopus ID, so we will accept any length.

### DIFF
--- a/schemas/core-2/metadata.json
+++ b/schemas/core-2/metadata.json
@@ -774,7 +774,7 @@
                       "Name_Identifier": {
                         "type": "string",
                         "title": "Identifier",
-                        "pattern": "[0-9]{11}"
+                        "pattern": "[0-9]+"
                       }
                     }
                   }
--- a/schemas/default-3/metadata.json
+++ b/schemas/default-3/metadata.json
@@ -1042,7 +1042,7 @@
                       "Name_Identifier": {
                         "type": "string",
                         "title": "Identifier",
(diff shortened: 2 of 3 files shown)


lwesterhof (approved):
LGTM!

PR #417, opened 2024-04-17
YDA-5659 - Modified policy to handle changing the role to already assigned role in the group.

kaur16:
YDA-5659 - Modified policy to handle changing the role to already assigned role in the group.


stsnel (approved):
LGTM! :+1:

PR #420, opened 2024-04-23
YDA-5663: lock down EUS use remove rule

stsnel:
It should only be called by (or on behalf of) a rodsadmin user.


lwesterhof (approved):
LGTM!

PR #421, opened 2024-04-25
Yda-5661 Do not put errors in logs for large files

claravox:


util/data_object.py
```diff
@@ -202,3 +209,13 @@ def decode_checksum(checksum):
         return "0"
     else:
         return binascii.hexlify(binascii.a2b_base64(checksum[5:])).decode("UTF-8")
+
+
+def get_groups(ctx, data_id):
```
stsnel: Can you add an integration test for this function to the data object integration tests, please? 

Existing integration tests:
https://github.com/UtrechtUniversity/yoda-ruleset/blob/1886415e0b17c6e96cbf4d279a2364544f873d7c/integration_tests.py#L60-L72

I feel it would be useful if the function name reflects what group role it is searching for. Perhaps something like `get_group_owners` or `get_groups_with_own_permission`?

revision_utils.py
```diff
@@ -1,21 +1,53 @@
...
+import hashlib
 import os
 
 from revision_strategies import get_revision_strategy
 from util import constants, log
 
 
-def calculate_end_of_calendar_day(ctx):
-    """Calculate the unix timestamp for the end of the current day (Same as start of next day).
+def revision_eligible(max_size, data_obj_exists, data_size, path, groups, revision_store_exists):
```
stsnel: Can we also incorporate the collection name and data object name checks in this function? They are currently in https://github.com/UtrechtUniversity/yoda-ruleset/blob/1886415e0b17c6e96cbf4d279a2364544f873d7c/revisions.py#L261-L266

The advantages of moving these checks here would be:
1. We have all the checks for seeing if a data object is eligible for revisions in one place.
2. We can add a unit tests or two for these checks
3. We can perform these checks both at AVU creation time and at revision creation time, so that we also cover situations where a data object has been moved / renamed after AVU creation, for example if a data object has been moved to trash shortly after being created.

stsnel (commented):
Looks good overall :+1: Also great that some of the long functions in revisions.py have now been decomposed in functions of a more manageable size.

Have added some comments about a few specific things above.

revisions.py
```diff
@@ -283,7 +287,10 @@ def resource_modified_post_revision(ctx, resource, zone, path):
                                                       pathutil.dirname(path), pathutil.basename(path), revision_avu_name),
                                                   offset=0, limit=1, output=genquery.AS_LIST))) > 0
 
-        if not already_has_avu:
+        # Check that size is not too big to make a revision
+        data_size = data_object.size(ctx, path)
+
+        if int(data_size) <= constants.UUMAXREVISIONSIZE and not already_has_avu:
```
stsnel: I think the size check can now be removed here, since it is also performed in `revision_eligible`, right?

claravox: Ah yes, thanks! I've corrected that now.

stsnel (approved):
Great, LGTM :+1:

PR #424, opened 2024-05-06
YDA-5696: harden rule_revision_batch

stsnel:
- Ensure it can only be started by rodsadmin users
- Validate balance ID and batch size parameters


lwesterhof (approved):
LGTM!

PR #430, opened 2024-06-03
YDA-5725

lwesterhof:


vault.py
```diff
@@ -251,7 +271,9 @@ def api_vault_unpreservable_files(ctx, coll, list_name):
 
     :returns: Set of unpreservable file formats
     """
-    zone = pathutil.info(coll)[1]
+    space, zone, _, _ = pathutil.info(coll)
+    if space is not pathutil.Space.VAULT:
```
stsnel: I believe we also need to permit research collections here, in order to facilitate checking for unpreservable file formats in the research module.

lwesterhof: Thanks, good catch.

PR #431, opened 2024-06-10
Yda 5393 copy to vault

claravox:
requires https://github.com/UtrechtUniversity/yoda/pull/427


claravox:
API tests pass locally. May have to do with the new configurations added

vault.py
```diff
@@ -828,24 +829,60 @@ def api_revoke_read_access_research_group(ctx, coll):
...
+    :returns: True for successful copy
     """
-    destination = target + '/original'
-    origin = folder
+    returncode = 0
+    try:
+        returncode = subprocess.call(["irsync", "-rK", "i:{}/".format(coll), "i:{}/original".format(target)])
+    except Exception as e:
+        log.write(ctx, "irsync failure: " + e)
+        log.write(ctx, "irsync failure for coll <{}> and target <{}>".format(coll, target))
```
claravox, line 909: I have never seen these lines run, even when clearly something went wrong because of wrong permissions on the folders.

claravox: This is expected, as an exception would only occur when something wrong went with subprocess itself.

lwesterhof (approved):
LGTM!

PR #432, opened 2024-06-12
YDA-5727: Script to update schema-id

claravox:
To run, for example:
`irule -r irods_rule_engine_plugin-python-instance -F tools/update-schema-id.r '*defaultSchema="'default-3'"'`

Tested locally by first deleting schema_id from some groups, and running this script to add them back.

### DIFF
--- a/tools/update-schema-id.r
+++ b/tools/update-schema-id.r
@@ -1,0 +1,102 @@
+#!/usr/bin/irule -r irods_rule_engine_plugin-python-instance -F
+#
+# Update the schema id for each group based on collection schema
+# 
+# Example command to run:
+# irule -r irods_rule_engine_plugin-python-instance -F tools/update-schema-id.r '*defaultSchema="'default-3'"'
+import genquery
+import session_vars
+
+def get_schema_collection(ctx, rods_zone, group_name, default_schema):
+    """Determine schema collection based upon rods zone and name of the group.
+    (shamelessly copied from schema.py, then tweaked)
+
+    If there is no schema id set on group level and
+    the category does not have a schema, 'default' is returned.
+
+    :param ctx:            Combined type of a callback and rei struct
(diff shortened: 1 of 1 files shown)


tools/update-schema-id.r
```diff
@@ -0,0 +1,100 @@
...
+    userIter = genquery.row_iterator(
+        "USER_GROUP_NAME",
+        "USER_TYPE = 'rodsgroup' AND USER_ZONE = '{}'".format(zone),
+        genquery.AS_LIST,
+        callback) 
+
+    for row in userIter:
+        name = row[0]
+        # Normalish groups
+        if name.startswith("research-") or name.startswith("deposit-") or name.startswith("grp-") or name.startswith("intake-"):
```
claravox: Does this get most of the groups we need?

lwesterhof: Yeah, but we can limit it to research. The grp group is legacy, intake does not use schemas and deposit is only one group in production.

tools/update-schema-id.r
```diff
@@ -0,0 +1,102 @@
...
+            schemaIdFound = False
+            for row1 in metaIter:
+                attr = row1[0]
+                if attr == 'schema_id':
+                    schemaIdFound = True
+            
+            if not schemaIdFound:
+                schema_collection = get_schema_collection(callback, zone, name, default_schema)
+                callback.uuGroupModify(name, 'schema_id', schema_collection, '', '') 
+                callback.writeLine("stdout", "Group {} set with schema id {}".format(name, schema_collection))
```
lwesterhof: Can we add a couple of extra log statements to show the start and finish of the process and also log groups that do have a schema id and don't need a change?

lwesterhof (changes requested):
Looks good, only some extra logging for the admin running the script would be nice.

PR #433, opened 2024-06-13
YDA-5726: improve ORCID transformation rule

stsnel:
Improve rule for transforming illformatted ORCID identifiers in multiple ways:
- Log more information, in particular print information about exceptions that occur, print any values that are transformed, and print values that cannot be transformed for verification and troubleshooting.
- Add a dry run mode
- Fix method for determining schema: use the schema identifier in the metadata file for this.
- Also transform default-3 and core-2 metadata, so that we have the option of running the ORCID transformation script after converting metadata to default-3/core-2.


lwesterhof (approved):
LGTM!

PR #434, opened 2024-06-14
YDA-5713: restrict access to anonymous account

stsnel:
Restrict access to the anonymous account using the iRODS protocol. Local access is always allowed. Remote access is only allowed if the network address has been explicitly permitted by the administrator.


stsnel:
Integration test fails are because the Yoda repo PR with the new remote address parameter hasn't been merged yet, and the new policy integration tests depend on that.

lwesterhof (approved):
LGTM

PR #437, opened 2024-06-17
YDA-5752: add vault metadata schema report

stsnel:
Add report tool for compiling a list of all vault data packages, their active metadata schema and whether the metadata matches the schema. This is meant to be used for verifying metadata upgrades and changes.


lwesterhof (approved):
Great, LGTM!

PR #446, opened 2024-06-26
YDA-5714: add security restrictions for GenQuery

stsnel:
This disallows queries for collection or data object AVUs without also selecting or adding a condition for a matching collection / data object column.


lwesterhof (approved):
This is great! LGTM

PR #447, opened 2024-06-27
YDA-5393: Retry copy to vault pending jobs too

claravox:
UI folder tests pass.
image

Related to https://github.com/UtrechtUniversity/yoda/pull/436


claravox:
While testing on development.yoda, ran into issue where revision or replication would start when copy to vault was trying to start as well. So the permissions on the folder might have already been changed by copy to vault, but then replication would try to run, and then copy to vault would never be run again

lwesterhof (approved):
LGTM!

PR #450, opened 2024-07-01
add RDFa to landing page

leonidastri:


templates/landingpage.html.j2
```diff
@@ -30,36 +30,36 @@
...
                         </div>
                         <section>
-                            <h2>{{ title }}</h2>
+                            <h2 property="dc:title">{{ title }}</h2>
                             <div class="meta">
                                 <label>Publication Date:</label>
-                                <span class="date">{{ publication_date }}</span>
-                                <label>Accessibility:</label>
+                                <span class="date" property="dc:created">{{ publication_date }}</span>
+                                <label property="dc:rights">Accessibility:</label>
```
lwesterhof: The property [accessRights](https://www.dublincore.org/specifications/dublin-core/dcmi-terms/terms/accessRights/) might be more appropriate here?

And the property should be on the span here?

templates/landingpage.html.j2
```diff
@@ -30,36 +30,36 @@
...
+                    <h1 property="dc:title">Data publication platform of Utrecht University</h1>
                 </div>
         </header>
         <main>
             <div>
                 <div class="row content">
                     <div class="col-lg-8">
                         <div class="creators">
-                            {% for creator in creators %} {{ creator.Name.Family_Name }}, {{ creator.Name.Given_Name }} {{ "&" if not loop.last }} {% endfor %}
+                            {% for creator in creators %} <span property="dc:creator"> {{ creator.Name.Family_Name }}, {{ creator.Name.Given_Name }} </span> {{ "&" if not loop.last }} {% endfor %}
```
lwesterhof: Maybe don't add the creator property here because we also add it later in the page?

templates/landingpage.html.j2
```diff
@@ -74,19 +74,19 @@
...
                                                     </td>
                                                 </tr>
                                             {% endfor %}
                                         </tbody>
                                     </table>
-                                    <small>
-                                        <b>Base DOI:</b>
-                                        <a href="https://doi.org/{{base_doi}}">https://doi.org/{{base_doi}}</a><br/>
-                                        This DOI represents all versions of this publication and will resolve to the latest publication.
+                                    <small typeof="dc:BibliographicResource">
```
lwesterhof: Not sure about the use of BibliographicResource here

templates/landingpage.html.j2
```diff
@@ -313,15 +313,15 @@
...
+                                <span property="dc:date">{{ publication_date }}</span>
                             </div>
                         </div>
                         <div class="row">
                             <div class="col-sm-2">
                                 <label>Last Modification</label>
                             </div>
                             <div class="col-sm-10">
-                                <span>{{ last_modified_date }}</span>
+                                <span property="dct:modified">{{ last_modified_date }}</span>
```
lwesterhof: These properties are from DCAT?

lwesterhof (changes requested):

templates/landingpage.html.j2
leonidastri: we can use the dc:modified from https://www.dublincore.org/specifications/dublin-core/dcmi-terms/

templates/landingpage.html.j2
leonidastri: Does this affect the other creator property? Because I have used the property title also more than once

lwesterhof (approved):
LGTM

PR #455, opened 2024-07-05
YDA-5783: fix logic for vault schemas

stsnel:
This fix implements additional logic for determining the active metadata schema of a vault group.

In addition to the existing behaviour (using the metadata schema of the matching deposit/research group), we now allow a schema_id AVU to be set on the vault group that overrides the metadata schema. Furthermore, if no vault schema_id AVU is set and no research/deposit group is found, we revert to the default metadata schema of the environment.

This ensures that metadata conversion can work on vault groups that no longer have a matching research/deposit group. If a standalone vault group has a non-default metadata schema, its metadata schema needs to be configured explicitly for it to be processed correctly by the schema transformation rules.

### DIFF
--- a/integration_tests.py
+++ b/integration_tests.py
@@ -14,6 +14,7 @@ import traceback
 import uuid
 
 import folder
+import schema
 from util import avu, collection, config, constants, data_object, log, msi, resource, rule, user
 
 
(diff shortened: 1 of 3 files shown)


lwesterhof (approved):
LGTM!

PR #456, opened 2024-07-05
Add tool for editing vault metadata

stsnel:


lwesterhof (changes requested):
Could we trigger the rule_meta_datamanager_vault_ingest to let the validation and provenance logging happen within the system, ensuring that all index jobs are executed as required?

stsnel:
I've now modified the script so that it uses the normal metadata ingest process for the vault by default.

One of the use cases for the script is to be able to bypass the normal metadata ingest process in cases where it can't be used (e.g. vault groups without a research group, categories without a datamanager group, etc.), so I've retained a command-line option for bypassing the normal process. The provenance log updates are now happening in the system though, even when we are bypassing the rest of the ingest process.

lwesterhof (approved):
LGTM!

PR #458, opened 2024-07-08
fix tests for data transfer page

leonidastri:
Just a quick fix for the tests because with the latest changes some of them did not work.

**Problem**
In the data_transfer.html file, the text "Gocommands" had been changed in "GoCommands".

**Solution**
Small change in the test_ui_data_transfer.py file, to search for "GoCommands" instead of "Gocommands"


lwesterhof (approved):
LGTM!

PR #460, opened 2024-07-09
YDA-5741: add users with keyboard ui tests

claravox:


claravox:
Now having some other failing UI tests that I think are unrelated.

PR #463, opened 2024-07-10
YDA 5718 add maintenance banner's UI Test

FuHsinyu:
1. Improve the scenario that when the user navigates to the admin page, "set maintenance banner" options are present.
3. Add a new scenario that the user sets up a banner message with/without marking it as important.
4. Add a new scenario that the user removes the existing banner.


FuHsinyu:
Awaiting for updating dumps of new testfile

lwesterhof (approved):
LGTM!

PR #465, opened 2024-07-11
Yda 5760 fix statistics

leonidastri:
https://utrechtuniversity.atlassian.net/browse/YDA-5760

**Description**
When user tries to sort groups by size in the Statistics module, the correct sorting order is not followed.

**Problem**
In api_resource_browse_group_data function of resources.py, data_size was a list of four integers. We sorted by size using the list of 4 integers.

**Solution**
Change the sorting mechanism in resources.py. In api_resource_browse_group_data function we change the sorting to use the last integer of the list


lwesterhof (approved):
LGTM!

PR #469, opened 2024-07-12
YDA-5809: fix scope schema transformations

stsnel:
Reduce the scope of collections that are processed for vault data package metadata transformations. Only metadata in the top vault data package collections should be considered for transformation.


lwesterhof (approved):
LGTM!

PR #470, opened 2024-07-15
yda-repair-script-unreg-change - Modified the subprocess call for unreg command when scope is 'replica'

kaur16:
yda-repair-script-unreg-change - Modified the subprocess call for unreg command when scope is 'replica'


stsnel (approved):
LGTM :+1:

PR #471, opened 2024-07-15
YDA-5810 fix rev cleanup handling removed data

stsnel:
Update revision logic to remove revisions of versioned data objects that no longer exist in the revision cleanup process.


lwesterhof (approved):
LGTM!

PR #475, opened 2024-07-18
Yda 5784 fix removing non existing avu

leonidastri:
**Problem**
In Yoda 1.9.3, removing an AVU from a user (and presumably other iRODS objects) using util.avu.rm_from_group results in an agent crash if the AVU does not exist.

 It is an iRODS problem, so we could not catch the error before iRODS shutting down the agent.

Bug description: https://utrechtuniversity.atlassian.net/browse/YDA-5784

**Solution**

Added a check to prevent the error from happening and therefore prevent iRODS from crashing


lwesterhof (commented):
LGTM!

PR #480, opened 2024-07-23
YDA-5835: fix zone selection in transform script

stsnel:
In the script for transforming publication metadata to support base DOIs, adjust the query condition for selecting the zone of a collection.

Selecting on USER_ZONE does not reliably return all data package collections in the zone. We need to either check the zone of the collection itself, or check the beginning of the collection name to select its zone.


lwesterhof (approved):
LGTM

PR #482, opened 2024-07-26
YDA-5598 - Removed the secured status and modified API and UI tests for folder status

kaur16:
YDA-5598 - Removed the secured status and modified API and UI tests for folder status


lwesterhof (approved):
LGTM

PR #484, opened 2024-08-01
YDA-5747: Script for granting read members access to vault packages

claravox:
Script to grant read members of research groups read access to vault packages that have already been published.

### DIFF
--- a/vault.py
+++ b/vault.py
@@ -1136,20 +1136,26 @@ def set_vault_permissions(ctx, coll, target):
     return True
 
 
-def reader_has_access(ctx, coll):
+def reader_needs_access(ctx, group_name, coll):
+    """Return if research group has access to this group but readers do not"""
     iter = genquery.row_iterator(
         "COLL_ACCESS_USER_ID",
         "COLL_NAME = '" + coll + "'",
         genquery.AS_LIST, ctx
     )
+    reader_found = False
+    research_found = False
+
     for row in iter:
         user_id = row[0]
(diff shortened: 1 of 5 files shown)


vault.py
```diff
@@ -1135,6 +1136,135 @@ def set_vault_permissions(ctx, coll, target):
...
+        if verbose:
+            modes.append("verbose")
+        log.write(ctx, "Running grant_readers_vault_access in {} mode.".format((" and ").join(modes)))
+
+    zone = user.zone(ctx)
+
+    # Get the group names
+    userIter = genquery.row_iterator(
+        "USER_GROUP_NAME",
+        "USER_TYPE = 'rodsgroup' AND USER_ZONE = '{}' AND USER_GROUP_NAME like 'research-%'".format(zone),
```
claravox, line 1251: Should this include more groups?

claravox:
The code isn't the most efficient, is that ok? Can either use more efficient algorithm or always set the vault permissions regardless of whether readers already have access.

vault.py
lwesterhof: No, only needed for research groups.

lwesterhof:
> The code isn't the most efficient, is that ok? Can either use more efficient algorithm or always set the vault permissions regardless of whether readers already have access.

Not a problem I think, we only need to run this once to fix the current inconsistencies.

lwesterhof (approved):
LGTM!

PR #485, opened 2024-08-01
YDA-5863: fix log messages schema transformation

stsnel:
The log messages for copying ACLs did not match the actions that the system was performing.


lwesterhof (approved):
LGTM!

PR #486, opened 2024-08-02
YDA-5864: ignore deleted users in schema transf.

stsnel:
In the schema transformation code, ignore ACLs of non-existent users when copying ACLs from a parent. This is needed because iRODS keeps ACLs referring to deleted users / groups around in the database (see https://github.com/irods/irods/issues/7778).


lwesterhof (approved):
LGTM!

PR #488, opened 2024-08-06
YDA-5871: limit returned groups in group data

stsnel:
When collecting group data for the group manager, collect only data for groups managed by the group manager. This prevents metadata on vault groups, in particular schema_id override AVUs (see YDA-5783), from breaking the group manager. More generally, it prevents the group manager from breaking because of group data for groups that are not managed by the group manager.


lwesterhof (approved):
LGTM!

PR #489, opened 2024-08-07
Yda-5720 UI tests for config publication terms

FuHsinyu:
Add UI tests for preview and update publication terms, associated to UI change in https://github.com/UtrechtUniversity/yoda-portal/pull/343

### DIFF
--- a/tests/features/ui/ui_admin.feature
+++ b/tests/features/ui/ui_admin.feature
@@ -2,9 +2,6 @@
 Feature: Admin UI
 
     Scenario Outline: Admin page view by admin users
-        technicaladmin is an admin user as role: rodsadmin
-        functionaladminpriv is an admin user in group: priv-admin
-
         Given user <user> is logged in
         When the user navigates to the admin page
         Then the text Administration is shown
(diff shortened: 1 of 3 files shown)


tests/features/ui/ui_admin.feature
```diff
@@ -73,3 +73,29 @@ Feature: Admin UI
...
+
+    Scenario Outline: Admin user updates publication terms
+        Given user <user> is logged in
+        When the user navigates to the admin page
+        And the user adds text <text> to publication terms
+        And the user clicks Update Terms button
+        And the user navigates to the admin page
+        Then the text <text> is displayed in the publication terms textarea
+        And the user removes the <text> from publication terms
+        And the user clicks Update Terms button
```
FuHsinyu, line 148: Here I removed the added text from publication terms for cleaning up, but I wonder is it better to have a separate test to do it?

lwesterhof: Yeah, that is cleaner. Also when it fails easier to determine where the error is.

PR #490, opened 2024-08-07
YDA-5808 - Modified to display Base DOI in System Metadata and all versions of selected datapackage in VAULT space.

kaur16:
YDA-5808 - Modified to display Base DOI in System Metadata and all versions of selected datapackage in VAULT space.

### DIFF
--- a/vault.py
+++ b/vault.py
@@ -563,7 +563,7 @@ def get_all_published_versions(ctx, path):
     if count == 1:  # Base DOI does not exist as it is first version of the publication
         # Convert the date into two formats for display and tooltip (Jan 1, 1990 and 1990-01-01 00:00:00)
         data_packages = [[x[0], datetime.strptime(x[1], "%Y-%m-%dT%H:%M:%S.%f").strftime("%b %d, %Y"), x[2],
-                            datetime.strptime(x[1], "%Y-%m-%dT%H:%M:%S.%f").strftime('%Y-%m-%d %H:%M:%S%z'), x[3]] for x in data_packages]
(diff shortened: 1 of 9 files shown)


vault.py
```diff
@@ -485,14 +485,19 @@ def api_vault_system_metadata(ctx, coll):
         landinpage_url = row[0]
         system_metadata["Landingpage"] = "<a href=\"{}\">{}</a>".format(landinpage_url, landinpage_url)
 
+    # Check for base DOI
```
lwesterhof: Lets discuss with the team, but maybe we can drop the DOIs from the system metadata if we always display the versions? Also when only a single version DOI exists?

vault.py
```diff
@@ -1467,17 +1512,18 @@ def get_approver(ctx, path):
         return None
 
 
-def get_doi(ctx, path):
+def get_doi(ctx, path, doi):
```
lwesterhof: Safer to add a default to doi: `doi="version"` and add a check to set it to version if it is not `base`?

```
def get_doi(ctx, path, doi="version"):
    if doi != "base":
        doi="version"
```

vault.py
```diff
@@ -580,6 +585,43 @@ def api_vault_collection_details(ctx, path):
 
     basename = pathutil.basename(path)
 
+    # If specific data package is selected, retreive all published versions
```
lwesterhof: Move this inside the "Check if collection is vault package" if else clause for a safer way to check this.

lwesterhof (changes requested):

vault.py
kaur16: Removed DOIs from system metadata section.

lwesterhof (approved):
LGTM!

PR #493, opened 2024-08-08
YDA-5865 - Added errors='replace' parameter in decode function to replace invalid UTF-8 sequences in file.

kaur16:
YDA-5865 - Added errors='replace' parameter in decode function to replace invalid UTF-8 sequences in file.


stsnel (approved):
LGTM! :+1: 

Can you just reformat the commit desciption so that the first line fits in the recommended 50 character limit, and any further lines in the 72 character limit before merging?

PR #499, opened 2024-08-20
Yda-5866 fix bug for revision creation failure when data object is removed

FuHsinyu:
Fixed the bug for following scenarios:
1. When no data object found -> skip the current revision
2. When data object exists -> return retrieved properties

Integration tests passed (in my local dev env)


integration_tests.py
```diff
@@ -518,6 +518,12 @@ def _test_folder_secure_func(ctx, func):
     {"name":   "util.data_object.exists.no",
      "test": lambda ctx: data_object.exists(ctx, "/tempZone/home/research-initial/testdata/doesnotexist.txt"),
      "check": lambda x: not x},
+    {"name":   "util.data_object.get_properties.default_properties",
+     "test": lambda ctx: data_object.get_properties(ctx, 12188, "irodsResc"),
+     "check": lambda x: x["DATA_NAME"] == "lorem.txt"},
+    {"name":   "util.data_object.get_properties.no_data_object",
+     "test": lambda ctx: data_object.get_properties(ctx, 1218812188, "irodsResc"),
```
FuHsinyu: what would be a "readable" non-exist-id? for now I use doubled data_id which surely not exist

util/data_object.py
```diff
@@ -24,6 +24,39 @@ def exists(ctx, path):
...
+    # Retrieve data obejct with default properties
+    query_fields = ", ".join(properties)
+    iter = genquery.row_iterator(
+        query_fields,
+        "DATA_ID = '{}' AND DATA_RESC_HIER like '{}%'".format(data_id, resource),
+        genquery.AS_LIST, ctx
+    )
+
+    # Return a None dict when no data object is found
+    prop_dict = {prop: None for prop in properties}
```
FuHsinyu: I made small change based on what we discussed. In case data object is not found, returns None dictionary rather than a None. This is because returning None will raise an error to stop the revision process (we expect to skip the current revision task). is that okay?

integration_tests.py
stsnel: If you want to be sure the data object ID does not exist, I think it's best to use an ID of an existing collection, resource or user. These iRODS entities use the same ID sequence as data objects, so if we have another type of iRODS entity with that ID, we can be sure it's not an ID of an existing data object.

It's probably best to add a comment about this in the integration test, since it may not be immediately obvious why this works.

util/data_object.py
stsnel: I think you can check the return value in `revision_create`, right?

I'm thinking of something along the lines of:
```
if revision_properties is None:
      log.write(ctx, "Not creating revision for data object with ID {}, because it no longer exists on resource {}.".format(data_id, resource))
      return False
```

revisions.py
```diff
@@ -563,7 +543,23 @@ def revision_create(ctx, print_verbose, data_id, resource, group_name, revision_
     :returns: True / False as an indication whether a revision was successfully created
     """
     revision_created = False
-    modify_time, data_size, coll_id, data_owner, basename, parent = get_data_object(ctx, data_id, resource)
+    log.write(ctx, "Revision_create")
```
stsnel: Debugging statement left here.

revisions.py
```diff
@@ -563,7 +543,23 @@ def revision_create(ctx, print_verbose, data_id, resource, group_name, revision_
...
+
+    # Retrive properties of the data object
```
stsnel: typo: Retrive -> Retrieve

integration_tests.py
stsnel: These integration tests looks like a good start :+1:  Can you also add a test / tests for the other properties? If it's hard to test a property exactly (e.g. modification time), I think just testing that something is returned should be good enough.

revisions.py
```diff
@@ -563,7 +543,23 @@ def revision_create(ctx, print_verbose, data_id, resource, group_name, revision_
...
+    modify_time = data_properties["DATA_MODIFY_TIME"]
+    data_size = data_properties["DATA_SIZE"]
+    coll_id = data_properties["COLL_ID"]
+    data_owner = data_properties["DATA_OWNER_NAME"]
+    basename = data_properties["DATA_NAME"]
+    parent = data_properties["COLL_NAME"]
+
+    # Skip current revision if data object is not found
+    if data_size is None:  # retrived data_size cannot be None
+        log.write(ctx, "ERROR - No data object found for data_id {}, move to the next revision creation".format(data_id))
```
stsnel: It would be useful to also log the resource, for troubleshooting purposes.

stsnel (changes requested):

integration_tests.py
FuHsinyu: Both are fixed!

util/data_object.py
FuHsinyu: I see what you mean, I have updated the codes per suggestions

stsnel (approved):
Great, looks good! :rocket:

PR #502, opened 2024-08-22
YDA-5801: copy user metadata from research to vault

lwesterhof:
Implements UtrechtUniversity/yoda#19

### DIFF
--- a/meta.py
+++ b/meta.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 """JSON metadata handling."""
 
-__copyright__ = 'Copyright (c) 2019-2023, Utrecht University'
+__copyright__ = 'Copyright (c) 2019-2024, Utrecht University'
 __license__   = 'GPLv3, see LICENSE'
 
 import json
@@ -26,8 +26,7 @@ __all__ = ['rule_meta_validate',
            'rule_meta_modified_post',
            'rule_meta_datamanager_vault_ingest',
            'rule_meta_collection_has_cloneable_metadata',
-           'rule_get_latest_vault_metadata_path',
-           'rule_copy_user_metadata']
+           'rule_get_latest_vault_metadata_path']
 
 
 def metadata_get_links(metadata):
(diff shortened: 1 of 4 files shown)


meta.py
```diff
@@ -682,31 +675,48 @@ def set_result(msg_short, msg_long):
...
-        for row in iter:
-            avu.associate_to_coll(ctx, target, row[0], row[1])
-
-        log.write(ctx, "rule_copy_user_metadata: copied user metadata from <{}> to <{}>".format(source, target))
+        # Retrieve all AVUs inside source collection.
+        user_metadata = list(avu.inside_coll(ctx, source, recursive=True))
+
+        # Group AVUs by entity and filter system metadata.
+        grouped_user_metadata = {}
+        for path, type, attiribute, value, unit in user_metadata:
```
claravox:
```suggestion
        for path, type, attribute, value, unit in user_metadata:
```

meta.py
```diff
@@ -682,31 +675,48 @@ def set_result(msg_short, msg_long):
...
+        grouped_user_metadata = {}
+        for path, type, attribute, value, unit in user_metadata:
+            if not attribute.startswith(constants.UUORGMETADATAPREFIX) and unit != constants.UUFLATINDEX and not unit.startswith(constants.UUUSERMETADATAROOT + '_'):
+                grouped_user_metadata.setdefault(path, {"type": type, "avus": []})
+                grouped_user_metadata[path]["avus"].append((attribute, value, unit))
+
+        # Generate metadata operations.
+        for path, item in grouped_user_metadata.items():
+            operations = {
+                "entity_name": path.replace(source, target + "/original"),
```
claravox: A bit far-fetched, but could there be a subfolder `path` with a second match of `source` so then the part would be replaced in `path` twice? For example:
```python3
source = '/tempZone/home/research-initial'
path = '/tempZone/home/research-initial/tempZone/home/research-initial/some-folder'
```

lwesterhof: Good catch!

util/avu.py
```diff
@@ -21,18 +21,70 @@
...
+    """
+    # coll+name -> path
+    def to_absolute(row, type):
+        if type == "collection":
+            return (row[1], type, row[2], row[3], row[4])
+        else:
+            return ('{}/{}'.format(row[0], row[1]), type, row[2], row[3], row[4])
+
+    collection_root = genquery.row_iterator(
+        "COLL_PARENT_NAME, COLL_NAME, META_COLL_ATTR_NAME, META_COLL_ATTR_VALUE, META_COLL_ATTR_UNITS",
```
claravox, line 61: I'm not familiar with what changes the speeds of genquery operations, or all the genquery quirks, but do we need to give back `COLL_PARENT_NAME` here?

lwesterhof: I don't think it has much impact on the speed because it is needed in the where condition.

Did it this way to be able to reuse to_absolute for all queries.

claravox (approved):
LGTM!

PR #508, opened 2024-08-26
YDA-5929: add parameter to update one publication

stsnel:
Add optional parameter to the update-publications rule so that it updates only a single publication, rather than all publications on the system. This is useful when troubleshooting an issue with a single publication.


lwesterhof (approved):
LGTM!

PR #509, opened 2024-08-26
YDA-5928: add logging update publication

stsnel:
Improve logging for updating publication endpoints, so that technical admins can see in the logs why the publication update rule is terminating early.


lwesterhof (approved):
LGTM!

PR #514, opened 2024-08-28
Yda 5889 add UI tests for non utf 8 name warning

leonidastri:
Add ui tests for uploading utf-8 characters filename and non utf-8 characters filename


lwesterhof (approved):
LGTM!

PR #515, opened 2024-08-28
YDA-5694 - Replaced msi_add_avu microservice with atomic metadata operations

kaur16:
YDA-5694 - Replaced msi_add_avu microservice with atomic metadata operations


lwesterhof (approved):
LGTM!

PR #516, opened 2024-08-28
Example for refactoring (part of) a legacy rule to Python

lwesterhof:


util/user.py
```diff
@@ -57,11 +57,17 @@ def from_str(ctx, s):
...
+    :param ctx:  Combined type of a callback and rei struct
+    :param user: Given user
+
+    :returns: Boolean indicating if user exists
+    """
     if type(user) is str:
         user = from_str(ctx, user)
 
-    return genquery.Query(ctx, "USER_TYPE", "USER_NAME = '{}' AND USER_ZONE = '{}'".format(*user)).first() is not None
+    return genquery.Query(ctx, "USER_TYPE", "USER_NAME = '{}' AND USER_ZONE = '{}'".format(*user)).first() in ["rodsuser", "rodsadmin"]
```
FuHsinyu, line 70: I am curious why only ('rodsuser' or 'rodsadmin') are considered user, what about other user types such as rodsgroup?

groups.py
```diff
@@ -705,6 +706,18 @@ def group_user_exists(ctx, group_name, username, include_readonly):
         return False
 
 
+@rule.make(inputs=[0], outputs=[1])
+def rule_user_exists(ctx, username):
+    """Rule wrapper to check if a user exists.
```
FuHsinyu, line 711: It'd be nice to also add "('rodsuser' or 'rodsadmin')" for better clarification in the doc strings

util/user.py
lwesterhof: In the context of the user utilities we only consider 'rodsuser' and 'rodsadmin' users. For 'rodsgroup' user types we have the group utilities.

FuHsinyu (commented):
Look generally good. Tested with multiple users and all were success. See inline comments for a few clarification.

stsnel (approved):
LGTM as well :+1:

PR #517, opened 2024-08-30
YDA-5829 troubleshooting tool for published data packages

FuHsinyu:
### New Shell Script
- **Added `troubleshoot_published_data.sh`**: This script troubleshoots packages based on user input. It provides options to either troubleshoot a single package or all published packages.

### New Python Rule
- **Task 1**: Retrieve all published data packages from the vault.
- **Task 2**: Conduct four types of verification for each published data package:
  - **Metadata Schema Check**: Ensures all data packages comply with the predefined metadata schema.
  - **AVU Check**: Checks for the presence and correctness of expected AVU pairs.
  - **Data Cite Registration**: Verifies registration with Data Cite through versionDOI and baseDOI.
  - **File Integrity Checks**: Checks the MD5 of landing page and combi JSON files against those on the public server.

### Pending Task
- ~~**Output to Terminal**: Implementing a feature to output results directly to the terminal (stdout) is pending and would enhance usability.~~ Completed!

### Continuation
- Sirjan and Claire may pick it up from here for changes due to my holiday leaves


claravox:
Would it be okay for `vault-metadata-schema-report.r` to print to standard out (or standard error) any errors instead of rodsLog?

troubleshoot_data.py
```diff
@@ -0,0 +1,432 @@
...
+            base_doi = get_val_for_attr_with_pub_prefix(ctx, data_package, "baseDOI")
+            status_code = datacite.metadata_get(ctx, base_doi)
+            base_doi_check = status_code == 200
+        except ValueError as e:
+            log.write(ctx, "check_datacite_doi_registration: Error while trying to get baseDOI - {}".format(e), write_stdout)
+
+    return (version_doi_check, base_doi_check)
+
+
+def get_val_for_attr_with_pub_prefix(ctx, data_package, attribute_suffix):
```
claravox: I'm not super happy with this name, happy for suggestions (or go back to `get_attribute_value`?).

troubleshoot_data.py
```diff
@@ -0,0 +1,434 @@
...
+        # Cannot check the metadata as technicaladmin
+        if not api_call:
+            schema_check_dict = vault_metadata_matches_schema(ctx, data_package, schema_cache, "troubleshoot-publications", write_stdout)
+            result['schema_check'] = schema_check_dict['match_schema'] if schema_check_dict else False
+
+        result['no_missing_AVUs_check'], result['no_unexpected_AVUs_check'] = check_print_data_package_system_avus(ctx, data_package, write_stdout)
+
+        # Do not check datacite in offline mode
+        if not offline:
+            result['versionDOI_check'], result['baseDOI_check'] = check_datacite_doi_registration(ctx, data_package, write_stdout)
```
claravox: Should we have a flag for the datacite doi check separate from the offline flag?

claravox:
It takes about 11 seconds to run this on development with 72 published packages. I can imagine it will be a lot slower on production.

util/misc.py
```diff
@@ -8,8 +8,83 @@
 import time
 from collections import OrderedDict
 
+import constants
 
-def last_run_time_acceptable(coll, found, last_run, config_backoff_time):
+
+def check_data_package_system_avus(extracted_avus):
```
claravox, line 14: Should we have another separate util avu file for just non-irods dependent functions?

troubleshoot_data.py
claravox: I added such a flag.

util/log.py
```diff
@@ -17,15 +17,19 @@
...
-    :param message: Message to write to log
+    :param ctx:          Combined type of a callback and rei struct
+    :param message:      Message to write to log
+    :param write_stdout: Whether to write to stdout (used for a few of our scripts)
     """
-    stack = inspect.stack()[1]
-    module = inspect.getmodule(stack[0])
-    _write(ctx, '[{}] {}'.format(module.__name__.replace("rules_uu.", ""), message))
+    if write_stdout:
+        ctx.writeLine("stdout", message)
```
lwesterhof, line 29: We do not want to include the originating module when writing to stdout?

troubleshoot_data.py
lwesterhof, line 1: Maybe rename this to publication_troubleshoot or  something else prefixed with publication to show that it is part of the publication module?

lwesterhof (changes requested):

util/log.py
claravox: The only other time we write to stdout this way as far as I can tell is for the update publications script. And it does not print the originating module either.

lwesterhof (approved):
LGTM!

PR #518, opened 2024-09-02
YDA-5939: Fix publication fail on third+ version

claravox:
Get baseDOIMinted from the previous version, use Python get rather than try except


lwesterhof (approved):
Nice, LGTM!

PR #519, opened 2024-09-04
Yda 5945 refactor update publications.r to python

leonidastri:
**Refactor**

Refactoring of update_publications.r from legacy code in r to python rule. The rule was moved inside the publication rules and has the same functionality as before.

Log messages from running the rule can be found in the iRods log.


lwesterhof (commented):
Nice, LGTM!

PR #522, opened 2024-09-12
Yda 5942 add user friendly messages csv import

leonidastri:
**Problem**

When modifying groups with a CSV, I use groups that are already created (to add users for example) . I check the “Allow updates”  button and “Process CSV”. I then get an error for existing groups even though the users have been added

**Solution**
When import groups using a CSV file, now the user can be informed with a simple, but useful descriptive message of the actions taken:

A message may consist of 1 or more of the following phrases:
- Group '<group_name>' created.
- Group '<group_name>' already exists.
- Users added (<no_users_added>).
- Users removed (<no_users_removed>).

**This should be merge taking into account the following PR also
https://github.com/UtrechtUniversity/yoda-portal/pull/362**

### DIFF
--- a/groups.py
+++ b/groups.py
@@ -988,7 +988,7 @@ def group_create(ctx, group_name, category, subcategory, schema_id, expiration_d
             if not sram.sram_connect_service_collaboration(ctx, short_name):
                 return api.Error('sram_error', 'Something went wrong connecting service to group "{}" in SRAM'.format(group_name))
 
-        if group.exists(ctx, group_name):
+        if group.exists(ctx, group_name) and not allow_update:
             return api.Error('group_exists', "Group {} not created, it already exists".format(group_name))
 
         response = ctx.uuGroupAdd(group_name, category, subcategory, schema_id, expiration_date, description, data_classification, co_identifier, '', '')['arguments']
--- a/groups.py
+++ b/groups.py
@@ -682,10 +682,10 @@ def apply_data(ctx, data, allow_update, delete_users):
         if user_added:
             message += ' User(s) added.'
         elif user_removed:
-            messages += ' User(s) removed.'
+            message += ' User(s) removed.'
         elif not new_group:
             message += ' No changes made.'
-            
+   
     return {"status": "ok", "message": message}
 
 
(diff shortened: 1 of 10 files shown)


uuGroup.r
```diff
@@ -776,7 +776,7 @@ uuGroupAdd(*groupName, *category, *subcategory, *schema_id, *expiration_date, *d
 	*kv."co_identifier"		  = *co_identifier;
 
 	# Shoot first, ask questions later.
-        *status = str(errorcode(msiSudoGroupAdd(*groupName, "manager", uuClientFullName, "", *kv)));
+ 	*status = str(errorcode(msiSudoGroupAdd(*groupName, "manager", uuClientFullName, "", *kv)));
```
stsnel: Can you remove this line from the PR, please, since it's unrelated?

util/api.py
```diff
@@ -30,8 +30,9 @@ def __init__(self, data=None, status='ok', info=None, debug_info=None):
         self.debug_info  = debug_info
 
     @staticmethod
-    def ok(**xs):
-        return Result(**xs)
+    def ok(info=None, **xs):
```
stsnel: I think this change is not necessary for the PR to work, right? In that case, can you remove this from the PR?

groups.py
```diff
@@ -978,7 +993,7 @@ def group_create(ctx, group_name, category, subcategory, schema_id, expiration_d
         message = response[9]
         if status == '0':
             return api.Result.ok()
-        elif status == '-1089000' or status == '-809000':
+        elif status == '-1089000' or status == '-809000' or status == '-806000':
```
stsnel, line 999: Good idea to add the check for SQL errors here.

On production environments we'd like to avoid SQL errors as much as possible, since they result in a significant amount of hard-to-read error messages in the logs. Can you also add a check to see if the group already exists just above the uuGroupAdd invocation? In that case, we should only get these SQL errors when two people are trying to create the same group simultanteously, which should be very rare.

stsnel (commented):
Looks good in general :+1:, I've added a few comments regarding some minor things

uuGroup.r
leonidastri: fixed

util/api.py
leonidastri: removed

groups.py
leonidastri: Done! Added check

leonidastri:
Resolved all issues! pr is ready!

PR #523, opened 2024-09-19
Yda 5892 delete revision creation avu for data objects found in trash

leonidastri:
The job for revisions now removes revision creation AVUs from data objects in trash.
This makes it **easier to monitor** the number of data objects waiting for revision creation.

**Additions:**
A function that removes revision creation AVUs from data objects found in the trash.

### DIFF
--- a/revisions.py
+++ b/revisions.py
@@ -361,7 +361,7 @@ def rule_revision_batch(ctx, verbose, balance_id_min, balance_id_max, batch_size
 
         minimum_timestamp = int(time.time() - config.async_revision_delay_time)
 
-        # Remove revision creation AVUs from deleted data objects. 
+        # Remove revision creation AVUs from deleted data objects.
         # This makes it easier to monitor the number of data objects waiting for revision creation.
         remove_revision_creation_avu_from_deleted_data_objects(ctx)
 
--- a/revisions.py
+++ b/revisions.py
@@ -363,7 +363,7 @@ def rule_revision_batch(ctx, verbose, balance_id_min, balance_id_max, batch_size
 
         # Remove revision creation AVUs from deleted data objects.
         # This makes it easier to monitor the number of data objects waiting for revision creation.
-        remove_revision_creation_avu_from_deleted_data_objects(ctx)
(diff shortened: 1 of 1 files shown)


revisions.py
```diff
@@ -1054,3 +1058,31 @@ def memory_limit_exceeded(rss_limit):
...
+        "COLL_NAME, DATA_NAME",
+        "COLL_NAME like '%{}/trash/home/%' AND META_DATA_ATTR_NAME = '{}'".format(user.zone(ctx), revision_avu_name),
+        genquery.AS_LIST, ctx
+    )
+
+    for coll_name, data_name in iter:
+        path = coll_name + '/' + data_name
+        try:
+            avu.rmw_from_data(ctx, path, revision_avu_name, "%")  # use wildcard cause rm_from_data causes problems
+            log.write(ctx, 'Removed revision creation AVUs from data object: {}'.format(path))
```
stsnel: Can we change this so that this message is only logged in verbose mode? In general, we try to log only errors and a summary when verbose mode is not enabled.

revisions.py
```diff
@@ -1054,3 +1058,31 @@ def memory_limit_exceeded(rss_limit):
...
+def remove_revision_creation_avu_from_deleted_data_objects(ctx):
+    """
+    Removes revision creation AVUs from deleted data objects [marked with 'org_revision_scheduled' metadata].
+
+    :param ctx:  Combined type of a callback and rei struct
+
+    :raises Exception:       If removal of revision creation AVUs fails
+    """
+    if user.user_type(ctx) != 'rodsadmin':
+        raise Exception("This rule can only be executed by a rodsadmin user.")
```
stsnel: These lines can be removed now that the function is no longer a rule.

stsnel (commented):
Looks good in general :+1:  ; I've added a few comments for specific things.

revisions.py
leonidastri: Change made! Now, it only logs the errors. If deletion of revision creation succeeds, it only logs the statement if verbose mode is enabled

leonidastri:
Made the necessary changes mentioned in the review! It's ready! 👍

stsnel:
Great, looks good :+1:

PR #524, opened 2024-09-23
Yda 5951 add transformation scopus isni

leonidastri:
The metadata schema transformation code from default-2 to default-3 currently handles ORCID-IDs and Researcher IDs, but not Scopus IDs and ISNI IDs. Because of this, metadata files with the default-2 schema containing Scopus or ISNI IDs can usually not be converted automatically to default-3. 

**Solution**
Add a transformation function that handles transformation of an ISNI and Scopus ID that consists of a series of digits (with optional spaces) to the format that is specified in the default-3 schema

### DIFF
--- a/schema_transformations.py
+++ b/schema_transformations.py
@@ -7,6 +7,7 @@ __license__   = 'GPLv3, see LICENSE'
 import re
 
 import meta
+from schema_transformations_utils import correctify_isni, correctify_orcid, correctify_researcher_id, correctify_scopus
 from util import *
 
 
@@ -133,7 +134,7 @@ def _default2_default3(ctx, m):
                     # Check for incorrect ORCID format.
                     if not re.search("^(https://orcid.org/)[0-9]{4}-[0-9]{4}-[0-9]{4}-[0-9]{3}[0-9X]$", person_identifier.get('Name_Identifier', None)):
                         corrected_orcid = correctify_orcid(person_identifier['Name_Identifier'])
-                        # Only it an actual correction took place change the value and mark this data as 'changed'.
+                        # Only if an actual correction took place change the value and mark this data as 'changed'.
                         if corrected_orcid is None:
                             log.write(ctx, "Warning: could not correct ORCID %s during schema transformation. It needs to be fixed manually."
                                       % (person_identifier['Name_Identifier']))
@@ -144,7 +145,7 @@ def _default2_default3(ctx, m):
                     # Check for incorrect Scopus format.
                     if not re.search("^\d{1,11}$", person_identifier.get('Name_Identifier', None)):
                         corrected_scopus = correctify_scopus(person_identifier['Name_Identifier'])
-                        # Only it an actual correction took place change the value and mark this data as 'changed'.
+                        # Only if an actual correction took place change the value and mark this data as 'changed'.
                         if corrected_scopus is None:
                             log.write(ctx, "Warning: could not correct Scopus %s during schema transformation. It needs to be fixed manually."
                                       % (person_identifier['Name_Identifier']))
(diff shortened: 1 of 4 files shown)


schema_transformations.py
```diff
@@ -138,6 +139,28 @@ def _default2_default3(ctx, m):
                                       % (person_identifier['Name_Identifier']))
                         elif corrected_orcid != person_identifier['Name_Identifier']:
                             person_identifier['Name_Identifier'] = corrected_orcid
+                # Check Scopus
+                elif person_identifier.get('Name_Identifier_Scheme', None) == 'Author identifier (Scopus)':
+                    # Check for incorrect Scopus format.
+                    if not re.search("^\d{1,11}$", person_identifier.get('Name_Identifier', None)):
+                        corrected_scopus = correctify_scopus(person_identifier['Name_Identifier'])
+                        # Only it an actual correction took place change the value and mark this data as 'changed'.
```
stsnel: Small typo here (it -> if)

schema_transformations.py
```diff
@@ -138,6 +139,28 @@ def _default2_default3(ctx, m):
...
+                            log.write(ctx, "Warning: could not correct Scopus %s during schema transformation. It needs to be fixed manually."
+                                      % (person_identifier['Name_Identifier']))
+                        elif corrected_scopus != person_identifier['Name_Identifier']:
+                            person_identifier['Name_Identifier'] = corrected_scopus
+                # Check ISNI
+                elif person_identifier.get('Name_Identifier_Scheme', None) == 'ISNI':
+                    # Check for incorrect ISNI format.
+                    if not re.search("^(https://isni.org/isni/)[0-9]{15}[0-9X]$", person_identifier.get('Name_Identifier', None)):
+                        corrected_isni = correctify_isni(person_identifier['Name_Identifier'])
+                        # Only it an actual correction took place change the value and mark this data as 'changed'.
```
stsnel: Small typo here (if -> it)

schema_transformations.py
```diff
@@ -722,6 +768,36 @@ def correctify_orcid(org_orcid):
...
 
 
+def correctify_scopus(org_scopus):
+    """Correct illformatted Scopus."""
+    # Get rid of all spaces.
+    new_scopus = org_scopus.replace(' ', '')
+
+    if not re.search("^\d{1,11}$", new_scopus):
+        # Return original value.
+        return org_scopus
```
stsnel: Don't we need to return None in this case, so that calling functions know that the value can't be transformed, and they can log a warning message? Can we also capture the behaviour of this function in a unit test?

schema_transformations.py
```diff
@@ -722,6 +768,36 @@ def correctify_orcid(org_orcid):
...
+
+    # Upper-case X.
+    new_isni = new_isni.replace('x', 'X')
+
+    # The last part should hold a valid id like eg: 123412341234123X.
+    # If not, it is impossible to correct it to the valid isni format
+    new_isni = new_isni.split('/')
+    if not re.search("^[0-9]{15}[0-9X]$", new_isni[-1]):
+        # Return original value.
+        return org_isni
```
stsnel: Don't we need to return None in this case, so that calling functions know that the value can't be transformed, and they can log a warning message? Can we also capture the behaviour of this function in a unit test?

stsnel (commented):
Looks good in general :+1: , I've added a few comments about specific parts.

schema_transformations.py
leonidastri: Moved correctify functions to schema_transformation_utils and and added unit tests for them

schema_transformations.py
leonidastri: Moved correctify functions to schema_transformation_utils and and added unit tests for them

stsnel:
Looks good to me :+1:

PR #530, opened 2024-10-23
YDA-5985: fix ARB update script for Ubuntu 20.04

stsnel:
The script needs to take into account that certificates are located in a different directory on Ubuntu systems.


lwesterhof (approved):
LGTM!

PR #534, opened 2024-12-05
CI: install any new portal dependencies if needed

stsnel:
Install any new portal dependencies in CI. These may be needed for running the API tests if the dependencies of the portal have changed since the Docker image was last built.


lwesterhof (approved):
LGTM!

PR #535, opened 2024-12-05
YDA-6040: add portal info API tests

stsnel:
On API test failures, print more information for troubleshooting purposes:
- CI: On failures, print web server error logs and portal access log for troubleshooting on portal level.
- API tests: if tests are unable to get a CSRF token, print the user that the error occurred with and the response from the web server for troubleshooting purposes.

(version for Yoda 1.10 and up)


lwesterhof (approved):
LGTM

PR #541, opened 2025-01-06
YDA-6082: specify vault resource on copy-to-vault

stsnel:
If a vault resource has been configured, use it when copying data from a research collection to the vault collection.


claravox (approved):
LGTM! I will/did leave you a comment about numThreads.

PR #542, opened 2025-01-08
YDA-6084: add sanity checks copy-to-vault

stsnel:
Add sanity checks for source and destination paths when copying data to the vault so that users can't make the vault module copy data in an unintended way by manipulating vault metadata. This also makes this part of the system more robust against unexpected inputs.


lwesterhof (approved):
LGTM!

PR #543, opened 2025-01-08
YDA-6059 fix datarequest api tests

FuHsinyu:
Root cause analysis: 
In irods, previously, the user can modify the ownership of the file without being the owner (the user may only have read-permission). but now, the user can only modify the ownership of the file, if they are the owner (or stronger rights like admin). For more discussion, see https://github.com/irods/irods/issues/6579 

What were done:
1. Failure in api_datarequest_submit and file_write_and_lock -> moved the current user to be the last who does permission or ownership modification
2. UPdated api.OK() call to api.Result.ok()

Note, in case you encountered the ReadTimeOut from HTTPConnectionPool, while running API tests. Try to update the timeout of HTTPConnectionPool, e.g, /yoda/yoda-ruleset/tests/venv_pytest/lib/python3.8/site-packages/urllib3/connectionpool.py in the function: _make_request()

Screenshot from 2025-01-08 15-08-15


lwesterhof (approved):
Great, LGTM!

PR #545, opened 2025-01-10
YDA-6104: display role changes on group import

stsnel:
When importing groups from a CSV file in the group manager, the portal did not show any role changes, which made it look like role changes were not processed.

This fix ensures that users can see any role changes that are made as a result of an import with the update flag enabled.


lwesterhof (approved):
LGTM!

PR #549, opened 2025-02-03
YDA-6067: Prevent create duplicate groups

claravox:
Prevent creation of overlapping research and deposit groups.


lwesterhof (approved):
LGTM!

PR #553, opened 2025-02-17
YDA-6198: add all users internal option

stsnel:
Add a way to make the ruleset consider all users to be internal, by setting the external domain filter to "*". This is useful in situations where all users are to be authenticated using OIDC.


lwesterhof (approved):
LGTM!

PR #554, opened 2025-02-19
YDA-6137 - Added list of pre-configured SRAM CO administrators as a parameter

kaur16:
YDA-6137 - Added list of pre-configured SRAM CO administrators as a parameter in POST request to SRAM in case of non-email users.


lwesterhof (approved):
LGTM!

PR #557, opened 2025-03-11
YDA-6244 fix encoding issue in ARB data manager

stsnel:
This resulted in, among other things, problems with uploading schemas in the playbook.


claravox:
This is for YDA-6182 right?

stsnel:
> This is for [YDA-6182](https://utrechtuniversity.atlassian.net/browse/YDA-6182) right?

Thanks, you're right. I'll close this PR and open a new one with the rightbranch name.

PR #559, opened 2025-03-17
YDA-4361: update schemas on different resources

stsnel:
This script makes it possible to reliably update a schema in Yoda from schema files in the local filesystem, in a way that can work with different replica situations. More specifically, the script can handle existing schema data objects that have no replicas inside the active resource tree.

### DIFF
--- a/tools/update-schema.py
+++ b/tools/update-schema.py
@@ -1,0 +1,133 @@
+#!/usr/bin/env python3
+""" Update schema in iRODS from local filesystem.
+"""
+
+import argparse
+import base64
+import hashlib
+import subprocess
+import os
+import re
+import sys
+from typing import List, Tuple, Union
+
+
+def parse_args() -> argparse.Namespace:
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument("schemadirectory", help="Schema directory on local filesystem")
+    parser.add_argument("schemacollection", help="Schema collection in iRODS")
+    parser.add_argument("--default-resource", default="irodsResc",
+                        help="Resource to put new data objects on")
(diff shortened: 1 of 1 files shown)


tools/update-schema.py
```diff
@@ -0,0 +1,133 @@
...
+
+def get_checksum(path, checksumtype):
+    if checksumtype == "md5":
+        hsh = hashlib.md5()
+    elif checksumtype == "sha2":
+        hsh = hashlib.sha256()
+    else:
+        raise ValueError(f"Checksum type {checksumtype} not supported.")
+
+    f = open(path, 'rb')
```
lwesterhof: Ensure that the file is properly closed after reading

```suggestion
    with open(path, 'rb') as f:
        while True:
            chunk = f.read(8192)
            if chunk:
                hsh.update(chunk)
            else:
                break
```

tools/update-schema.py
```diff
@@ -0,0 +1,133 @@
...
+    sys.exit(1 if anything_failed else 0)
+
+
+def data_object_exists(collection_name: str, dataobject_name: str) -> bool:
+    checkexist_command = ["ils", os.path.join(collection_name, dataobject_name)]
+    (checkexist_code, checkexist_stdout, checkexist_stderr) = run_command(checkexist_command)
+    return "does not exist" not in checkexist_stderr
+
+
+def get_checksum(path, checksumtype):
```
lwesterhof: Return and argument types str?

lwesterhof (changes requested):

lwesterhof (approved):
Nice improvement, LGTM!

PR #560, opened 2025-03-18
YDA-6243: Fix el9 ca_file permission issue when running arb-update-resources.py

leonidastri:
This is a fix for el9 ca_file permission issue when running arb-update-resources.py


lwesterhof (approved):
LGTM!

PR #563, opened 2025-03-26
YDA-6182 Fix ARB data encoding issue

stsnel:
Fix ARB encoding issue after upgrade to Python 3. This caused schema data objects to not be uploaded to iRODS initially.


lwesterhof (approved):
LGTM!

PR #564, opened 2025-03-27
YDA-6197 - Added UU function as a workaround for msiArchiveCreate microservice

kaur16:
YDA-6197 - Added UU function as a workaround for msiArchiveCreate microservice


lwesterhof (approved):
LGTM!

PR #565, opened 2025-03-31
Yda 6068 add csv output to troubleshoot tool

FuHsinyu:
1. Run in human readable output (by default):
`python3 troubleshoot-published-data.py -m human` or `python3 troubleshoot-published-data.py`
2. Run in CSV output:
`python3 troubleshoot-published-data.py -m csv` 

Documentation update: https://github.com/UtrechtUniversity/yoda/pull/551

### DIFF
--- a/publication_troubleshoot.py
+++ b/publication_troubleshoot.py
@@ -208,7 +208,7 @@ def compare_local_remote_landingpage(ctx, file_path, url, offline, api_call, wri
     # Set encoding to utf-8 for the response text (otherwise will not match local_data)
     # response.text is then returned as unicode
     response.encoding = 'utf-8'
-    local_data_uni = local_data.decode("utf-8")
+    local_data_uni = local_data #.decode("utf-8") # TODO: This causes error
 
     if local_data_uni == response.text:
         return True
--- a/publication_troubleshoot.py
+++ b/publication_troubleshoot.py
@@ -331,19 +331,21 @@ def batch_troubleshoot_published_data_packages(ctx, requested_package, log_file,
     :returns: A dictionary of dictionaries providing the results of the job.
     """
 
+    # Only write_stdout when in human mode and api_call is not enabled
     write_stdout = not api_call and (mode == 'human')
     # Check permissions - rodsadmin only
     if user.user_type(ctx) != 'rodsadmin':
         log.write(ctx, "User is not rodsadmin", write_stdout)
         return {}
 
+    # Collect data packages
     data_packages = collect_troubleshoot_data_packages(ctx, requested_package, write_stdout)
     if not data_packages:
         return {}
     schema_cache = {}
     results = {}
 
-    # Troubleshooting steps
+    # Troubleshooting packages
     for data_package in data_packages:
         result = {}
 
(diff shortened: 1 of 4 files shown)


FuHsinyu:
Example output:
`irods@combined:/etc/irods/yoda-ruleset/tools$ python3 troubleshoot-published-data.py -o -l -m csv`

`
Package,Schema Check,Missing AVUs Check,Unexpected AVUs Check,Version DOI Check,Base DOI Check,Landing Page Check,Combi JSON Check
/tempZone/home/vault-initial/testdata[1742823456],Pass,Fail,Pass,Fail,Fail,Pass,Pass
`

I am not sure if it is right to have these Fails... I'll have a look later if it is my package issue. But the PR is ready for review.

FuHsinyu:
Just checked the API test which failed. The reason is that we added new AVU to data package, in my failed case that is `check_data_package_system_avus: There are some missing AVUs in data package </tempZone/home/vault-initial/testdata[1742823456]> - ['org_publication_versionDOIMinted', 'org_publication_landingPageUploaded', 'org_publication_anonymousAccess', 'org_publication_oaiUploaded', 'org_publication_submission_actor']`

Shall we add them to the ground truth? This manual operation seems to be a issue in future as well, how do we improve it, if you have any suggestions

FuHsinyu:
Currently, there is an error when you run the troubleshooting tool with python3, because we upgraded python2 to python 3. See the code `local_data_uni = local_data.decode("utf-8")`. A workaround is to use `local_data_uni = local_data#.decode("utf-8")` in publication-troubleshoot.py

FuHsinyu:
Currently, a potential improvement is that: when outputting a failed package with a failed reason (e.g, missing AVUs), in human mode, the reason will be displayed at the top of the output. Rather, I wonder if it be nice to have the reason combined with the package output. Let me know what you think.
e.g, 
Troubleshooting package No.1:
Missing AVU found...
Troubleshooting package No.2:
MIssing landing page.

Results for: NO.1
Package FAILED one or more tests:
Schema Check: Pass
Missing AVUs Check: Fail
Unexpected AVUs Check: Pass
Version DOI Check: Fail
Landing Page Check: Pass
Combi JSON Check: Pass

Results for: NO.2
Package FAILED one or more tests:
Schema Check: Pass
Missing AVUs Check: Pass
Unexpected AVUs Check: Pass
Version DOI Check: Fail
Landing Page Check: Fail
Combi JSON Check: Pass

publication_troubleshoot.py
```diff
@@ -398,7 +378,15 @@ def batch_troubleshoot_published_data_packages(ctx, requested_package, log_file,
                 json.dump(result, writer)
                 writer.writelines('\n')
 
-    return results
+    # Convert booleans to Pass/Fail strings
+    converted_results = {}
+    for pkg, res in results.items():
+        converted_results[pkg] = {
+            key: 'Pass' if value else 'Fail' if isinstance(value, bool) else value
```
stsnel: Are we sure these values are boolean? If so, is it possible to simplify this expression to something like `'Pass' if value else 'Fail'`? If not, perhaps it would be good to add a comment about the intent of this expression.

stsnel (commented):
This looks good in general :+1:! Can you remove trailing whitespace in `troubleshoot_data.r` and have a look at the comment above?

publication_troubleshoot.py
FuHsinyu: You are right. These values can only be Boolean. I have updated the codes accordingly.

stsnel (approved):
LGTM :+1:

PR #566, opened 2025-04-04
YDA-6231

lwesterhof:


folder.py
```diff
@@ -415,15 +420,16 @@ def folder_secure_succeed_avus(ctx: rule.Context, coll: str, group_name: str) ->
...
+            return False
 
     # Remove target AVU on source folder. This should be done after all possibly failing steps
     # have occurred in folder_secure (any "return False" steps), so that if those trip a retry state,
     # on retry folder_secure can reuse the target from before.
-    if (not group_name.startswith("deposit-")
-            and not avu.rmw_from_coll(ctx, coll, constants.IICOPYPARAMSNAME, "%", True)):
-        return False
+    if not group_name.startswith("deposit-") and constants.IICOPYPARAMSNAME in org_metadata:
+        if not avu.rmw_from_coll(ctx, coll, constants.IICOPYPARAMSNAME, "%", True):
```
claravox, line 431: I was wondering if we could merge some of the nested if statements in this PR, or if there is a reason to keep them this way.

lwesterhof: Yeah, fair point. The reason I did not do it in this case is that in the first condition nothing is executed, but in the second condition something is executed. So it might be a bit easier to read?

claravox (approved):
LGTM! Although I did not do particularly thorough testing it makes sense to me.

PR #569, opened 2025-04-08
Update logging functions from iRODS 4.3

stsnel:
- No longer print user/zone information in log function, since printing user/zone information is now handled by the new log transform script.
- Also make all other parts of the code use the public log.write function rather than calling log._write directly in some places.


lwesterhof (approved):
LGTM!

PR #570, opened 2025-04-14
Add parameter for measuring coverage to ruleset

stsnel:
This enables us to measure coverage during testing; we can use this to integrate coverage measurements into our CI workflow for integration and API tests later.


lwesterhof (approved):
Nice, LGTM!

PR #573, opened 2025-04-22
Update schema transformation rules -> decorator

stsnel:
Update schema transformation rules to use the default rule decorator, in order to improve readability and enable type checking.


lwesterhof (approved):
Nice, LGTM!

PR #575, opened 2025-05-12
Fix publication update AVU in DM metadata ingest

stsnel:
With the workaround for https://github.com/irods/irods/issues/8265, the publication update AVU in the data manager metadata ingest function was changed inadvertently. This resulted in publication updates not being picked up by the publication processing job anymore. This commit restores the correct publication update AVU.


kaur16:
Looks good to me!

PR #576, opened 2025-05-14
YDA-6344: block collection mounting

stsnel:
Block mounting and unmounting collections, as well as creating soft collection links and synchronizing mounted collections.

This is disabled as part of application hardening, since these operations are not used in any legitimate application feature and involve potential security risks.


lwesterhof (approved):
LGTM!

PR #582, opened 2025-06-04
YDA6364: ensure path is in a valid space for research APIs

lwesterhof:


stsnel (commented):
I have two questions / comments:
- The temporary file list and research manifest endpoints will now fail silently if they are used outside of approved spaces. Can we make them fail explicitly using an API error to make it clear that the endpoint could not return results due to an error condition?
- Can we add integration tests for this behaviour?

stsnel (approved):
LGTM :+1: !

PR #583, opened 2025-06-06
Release of version 2.0 of hashes_collection.sh script

leonidastri:
Update of the hashes_collection.sh script to version 2.0. This is related to the ticket [YDA-6378](https://utrechtuniversity.atlassian.net/browse/YDA-6378)

The prior release of the script did not account for empty collections.

**Changes since release of  version 1.0:**
- Output in 'report' format, includes context, checksum, and a list of objects
- Reports separate counts for number of data objects and number of subcollections
- Hash is **performed on all objects, including empty collections**
- Only **good replicas** are considered in calculation of the hash value
- The script uses temporary files to store intermediate results, in a configurable directory
- Addition of **--detailed** parameter for detailed output
- Addition of **--debug** parameter for keeping temporary files for debugging purposes

### DIFF
--- a/tools/hashes_collection.sh
+++ b/tools/hashes_collection.sh
@@ -1,18 +1,116 @@
 #!/bin/bash
-# set of shell commands to hash the set of data objects and hierarchy in a collection
-# To compare the content of 2 collections, update the shell variable COLLECTION and rerun the query line
-# explanation:
-# 1. we select logical path and data object size for each data object within the collection tree
-# 2. we remove the prefix (collection name) so that the logical name starts within the collection (needed for comparison)
-# 3. we sort the pathnames, using a locale that ensures all characters have a different sorting order
-# bash commands:
 
-# Check if COLLECTION argument is provided
-if [ -z "$1" ]; then
-    echo "Error: Collection path must be provided as an argument." >&2
-    echo "Usage: $0 <collection-path>" >&2
+# Exit immediately if a command exits with a non-zero status.
+set -e
+
+#Set of shell commands to hash the set of data objects and hierarchy in a collection
+#to compare the content of 2 collections, update the shell variable COLLECTION and rerun the query line
+
+#explanation:
+#1. we select logical path and data object size for each data object within the collection tree
+#2. only good replicas (status '1') of data objects are considered for the hash calculation.
+#3. we select logical path for each subcollection within the collection tree
+#4. we remove the prefix (collection name) so that the logical name starts within the collection (needed for comparison)
+#5. we sort the pathnames (data objects and subcollections), using a locale that ensures all characters have a different sorting order
+
+DETAILED=0
+DEBUG=0
+
+# Define the hash calculation version
+VERSION="v2.0"
+
+# Initialize COLLECTION to be empty. This script strictly expects COLLECTION
+# to be provided as a command-line argument.
+COLLECTION=""
+
+# Parse arguments
+for arg in "$@"; do
+    case "$arg" in
+        --detailed) DETAILED=1 ;;
+        --debug) DEBUG=1 ;;
+        # The first positional argument found is taken as the COLLECTION.
+        # This will override any existing environment variable for the script's internal use.
+        *) COLLECTION="${arg%/}" ;;
+    esac
+done
+
+# If COLLECTION is still empty after parsing arguments, it means it wasn't provided.
+if [ -z "$COLLECTION" ]; then
+   echo "Usage: bash $0 <collectionName> [--detailed] [--debug]" >&2
+   echo "ERROR: Missing required collection name argument." >&2
+   exit 1
+fi
+
+# Using ils -d to check for collection existence
+if ! ils -d "${COLLECTION}" > /dev/null 2>&1; then
+    echo "ERROR: iRODS collection '${COLLECTION}' does not exist or is inaccessible." >&2
+    echo "Please verify the collection path and your iRODS environment." >&2
     exit 1
 fi
 
-COLLECTION="${1%/}/"
-iquest --no-page "%s/%s %s" "select COLL_NAME,DATA_NAME,DATA_SIZE where COLL_NAME like '${COLLECTION}%'"|cut -c$((${#COLLECTION}+1))-|LC_ALL=C sort|sha256sum
+TEMPFILE_DATE=$(date +%Y%m%d)
+TEMPFILE="$(mktemp --tmpdir=/tmp "check_objects.${TEMPFILE_DATE}.XXXXX")" || { echo "ERROR: Failed to create temporary file for objects." >&2; exit 1; }
+TEMPFOLDER="$(mktemp --tmpdir=/tmp "check_collections.${TEMPFILE_DATE}.XXXXX")" || { echo "ERROR: Failed to create temporary file for collections." >&2; exit 1; }
+TEMPSORTED="$(mktemp --tmpdir=/tmp "check_sorted.${TEMPFILE_DATE}.XXXXX")" || { echo "ERROR: Failed to create temporary file for sorted output." >&2; exit 1; }
+
+export TEMPFILE
+export TEMPFOLDER
+export TEMPSORTED
+
+# Ensure temporary files are cleaned up on exit, unless DEBUG is enabled.
+if [ "$DEBUG" -ne 1 ]; then
+    trap 'rm -f "$TEMPFILE" "$TEMPFOLDER" "$TEMPSORTED"' EXIT
+fi
+
+# --- Query 1: Data objects within subcollections (good replicas)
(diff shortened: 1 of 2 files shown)


leonidastri:
Some more testing that everything works fine is required before merge.

leonidastri:
Testing is finished. It can be reviewed

tools/hashes_collection.sh
```diff
@@ -1,18 +1,60 @@
...
-if [ -z "$1" ]; then
-    echo "Error: Collection path must be provided as an argument." >&2
-    echo "Usage: $0 <collection-path>" >&2
-    exit 1
+
+#set of shell commands to hash the set of data objects and hierarchy in a collection
+#To compare the content of 2 collections, update the shell variable COLLECTION and rerun the query line
+#explanation:
+#1. we select logical path and data object size for each data object within the collection tree
+#2. we remove the prefix (collection name) so that the logical name starts within the collection (needed for comparison)
```
stsnel: The explanation does not mention subcollection names, whereas the queries below do take these into account. I think either this step needs to be added to the explanation, or (alternatively) we can remove the `explanation` part of the comments.

tools/hashes_collection.sh
```diff
@@ -1,18 +1,60 @@
...
+done
+
+if [ -z "$COLLECTION" ]; then
+   echo "Usage: bash $0 <collectionName> [--detailed] [--debug]"
+   exit 1
 fi
 
-COLLECTION="${1%/}/"
-iquest --no-page "%s/%s %s" "select COLL_NAME,DATA_NAME,DATA_SIZE where COLL_NAME like '${COLLECTION}%'"|cut -c$((${#COLLECTION}+1))-|LC_ALL=C sort|sha256sum
+DIR="/tmp"
```
stsnel: If the script uses static temporary file paths like this, running multiple instances of the script simultaneously may result in unpredictable errors, since the instances of the script will be overwriting each other's temporary files. We need to use `mktemp` here, or a similar way to deconflict temporary files of multiple instances.

tools/hashes_collection.sh
```diff
@@ -1,18 +1,60 @@
...
+
+if [ "$DETAILED" -eq 1 ]; then
+  echo "DATA MANIFEST"
+  echo "-------------"
+  echo "Manifest created on : $(date -u)"
+  echo "Content of directory: ${COLLECTION_NO_SLASH}"
+  echo "File count          : $(wc -l <${TEMPFILE})"
+  echo "Subdirectories count: $(wc -l <${TEMPFOLDER})"
+  echo "Checksum            : ${CHECKSUM}"
+  echo "-------------"
```
stsnel, line 105: This PR changes the way hashes are calculated in a non-backwards compatible way. That's okay, but I suggest adding a version number for the hash calculation method to the output so that operations staff can see in the output whether the hashes they would like to compare were calculated using a compatible method. I guess it would be needed in both the detailed and default output.

tools/hashes_collection.sh
```diff
@@ -1,18 +1,60 @@
...
 fi
 
-COLLECTION="${1%/}/"
-iquest --no-page "%s/%s %s" "select COLL_NAME,DATA_NAME,DATA_SIZE where COLL_NAME like '${COLLECTION}%'"|cut -c$((${#COLLECTION}+1))-|LC_ALL=C sort|sha256sum
+DIR="/tmp"
+export TEMPFILE=${DIR}/check_objects.tmp
+export TEMPFOLDER=${DIR}/check_collections.tmp
+export TEMPSORTED=${DIR}/check_sorted.tmp
+
+iquest --no-page "%s/%s %s" "select COLL_NAME,DATA_NAME,DATA_SIZE where COLL_NAME like '${COLLECTION}/%'" | grep -v '^CAT_NO_ROWS_FOUND' | cut -c$((${#COLLECTION}+1))- >${TEMPFILE}
```
stsnel: This query, as well as the one below, also compares non-good (e.g. `stale` or `intermediate`) replicas if they have different sizes. That's probably not what we want as default behaviour. I suggest adding a `DATA_REPL_STATUS` condition to these queries. If we decide to keep it out of scope of the present ticket,  we need to at least document this as a limitation of the current version of the script.

It would also be good to check the return code of the `iquest` queries here, so that we can return an error message if something goes wrong with the queries, rather than a checksum that is not based on the real contents of the collection. We can't use `set -o pipefail` here, because `CAT_NO_ROWS_FOUND` with return code `1` is a valid result, but we can check for codes other than 0/1 using `PIPESTATUS`

tools/hashes_collection.sh
```diff
@@ -1,18 +1,60 @@
...
+  echo " "
+  echo "List of Subdirectories and Files (size in bytes)"
+  echo "================================================"
+  cat ${TEMPSORTED}
+else
+  echo "${CHECKSUM}"
+fi
+
+if [ "$DEBUG" -ne 1 ]; then
+  rm -f ${TEMPFILE} ${TEMPFOLDER} ${TEMPSORTED}
```
stsnel: Please quote arguments here to prevent argument splitting if any of these names are changed to e.g. contain whitespace.

tools/hashes_collection.sh
```diff
@@ -1,18 +1,60 @@
...
+        --detailed) DETAILED=1 ;;
+        --debug) DEBUG=1 ;;
+        *) if [[ -z "$COLLECTION" ]]; then COLLECTION="${arg%/}"; fi ;;
+    esac
+done
+
+if [ -z "$COLLECTION" ]; then
+   echo "Usage: bash $0 <collectionName> [--detailed] [--debug]"
+   exit 1
 fi
```
stsnel, line 43: Can we add a check here so that the command fails with an error message when used with a non-existent `COLLECTION`? This makes it easier for users to see what's going on if checksums don't match because e.g. there was a typo in one of the collection names.

tools/hashes_collection.sh
```diff
@@ -1,18 +1,60 @@
...
+
+#set of shell commands to hash the set of data objects and hierarchy in a collection
+#To compare the content of 2 collections, update the shell variable COLLECTION and rerun the query line
+#explanation:
+#1. we select logical path and data object size for each data object within the collection tree
+#2. we remove the prefix (collection name) so that the logical name starts within the collection (needed for comparison)
+#3. we sort the pathnames, using a locale that ensures all characters have a different sorting order
+
+#bash commands:
+
```
stsnel, line 15: Adding a `set -e` here would make the script stop in case of unexpected failures, rather than continue and return a potentially wrong checksum.

tools/hashes_collection.sh
```diff
@@ -1,18 +1,60 @@
...
+for arg in "$@"; do
+    case "$arg" in
+        --detailed) DETAILED=1 ;;
+        --debug) DEBUG=1 ;;
+        *) if [[ -z "$COLLECTION" ]]; then COLLECTION="${arg%/}"; fi ;;
+    esac
+done
+
+if [ -z "$COLLECTION" ]; then
+   echo "Usage: bash $0 <collectionName> [--detailed] [--debug]"
```
stsnel: The current version of the script silently ignores the collection name argument if a `COLLECTION` variable is present in the environment. Does this behaviour facilitate a particular use case? If so, I suggest making it clear that the argument is being ignored by e.g. printing a message that the collection name from the environment is used instead and the command-line argument is ignored. If we don't have a specific reason for this behaviour, we can simplify by just always using the command-line argument.

stsnel (commented):
I think the main idea of this change is good. I've added a few comments below about some specific aspects of the implementation that I believe can still be improved.

Can you also describe the reason for this change and mention the ticket number in the commit text, please? It can be useful for future developers who have questions about this script to have a bit of context here.

tools/hashes_collection.sh
leonidastri: Changed as suggested.

tools/hashes_collection.sh
leonidastri: Changed as suggested.

tools/hashes_collection.sh
leonidastri: Changed as suggested.

tools/hashes_collection.sh
leonidastri: Changed as suggested.

tools/hashes_collection.sh
leonidastri: Changed as suggested.

tools/hashes_collection.sh
leonidastri: Changed as suggested.

tools/hashes_collection.sh
leonidastri: Changed as suggested.

tools/hashes_collection.sh
leonidastri: Changed as suggested.

stsnel (approved):
Great, looks good to me! :+1:

PR #587, opened 2025-06-11
Add Lookup for "open" invitations

HarryKodden:
Added check to lookup for any existing invitations for the collabaration + user with status 'open'
This prevents adding new invitations when there are already invitation outstanding.

*** Disclaimer: Please treat as psuedo/untested code  ***

### DIFF
--- a/sram.py
+++ b/sram.py
@@ -158,7 +158,7 @@ def sram_put_collaboration_invitation(ctx: rule.Context, group_name: str, userna
     url = "{}/api/invitations/v1/invitations/{}".format(config.sram_rest_api_url, co_identifier)
 
     if config.sram_verbose_logging:
-        log.write(ctx, "get {}: {}".format(url, response.status_code))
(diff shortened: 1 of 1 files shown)


sram.py
```diff
@@ -152,9 +152,30 @@ def sram_put_collaboration_invitation(ctx: rule.Context, group_name: str, userna
...
+
+    if config.sram_verbose_logging:
+        log.write(ctx, "get: {}".format(url))
+
+    response = requests.get(url, headers=headers, timeout=30, verify=config.sram_tls_verify)
+
+    if config.sram_verbose_logging:
+        log.write(ctx, "response: {}".format(response.status_code))
+
+    if response.status != 200:
```
kaur16: Should be 'status_code'

kaur16:
Looks good to me!

lwesterhof (approved):
LGTM

PR #588, opened 2025-06-16
Fix: missing log argument in publication status check

claudisa-uu:
Fixing a missing argument to log.write() call in check publication status function


lwesterhof (approved):
LGTM, good catch!

PR #596, opened 2025-07-01
YDA-6336: add specific API error for no access permission

claudisa-uu:
This PR adds a more specific API error for the case of no access permission on metadata file.


lwesterhof (approved):
LGTM!

PR #605, opened 2025-08-07
YDA-6473 - Added support for depublished data packages in update publications rule

kaur16:
YDA-6473 - Added support for depublished data packages in update publications rule


lwesterhof (approved):
LGTM!

PR #610, opened 2025-08-21
yda-6460: add script to archive data collection

FuHsinyu:
Added a script to archive and extract files within irods. If the archive extension is .zip, archives are also compressed.  
Limitations: currently, target resource is set to be the default, specifying files for extraction is not supported, in addition .zip extraction is not supported due to msiArchiveExtract limitations.


tools/archive_data_collection.sh
```diff
@@ -0,0 +1,166 @@
...
+
+# Capture rule execution output and time the operation
+start_time=$(date +%s)
+
+# Start elapsed time counter in background
+elapsed_time "$start_time" &
+timer_pid=$!
+
+# Execute rule and capture output
+/bin/irule -r "$DEFAULT_RULE_ENGINE" -F "$RULE_FILE" \
```
lwesterhof: Could we consider adding the rule as a string directly within this script? This way, we can keep all related information consolidated in a single file?

lwesterhof (changes requested):

tools/archive_extract_data.sh
```diff
@@ -0,0 +1,204 @@
...
+EOF
+            errors=$((errors+1))
+        fi
+    elif [[ "$MODE" == "extract" ]]; then
+        # Check for .zip extension (case-insensitive)
+        if [[ "${SOURCE_PATH,,}" == *.zip ]]; then
+            cat <<EOF >&2
+
+ERROR: ZIP archive extraction is not supported for now.
+Please use your local applications to extract such as unzip cmd (Linux/Mac), 7-Zip/WinZip (Windows)
```
claravox:
```suggestion
Please download the archive and use a local application to extract such as unzip cmd (Linux/Mac), 7-Zip/WinZip (Windows)
```

claravox (approved):
LGTM!

PR #613, opened 2025-08-28
Refactor package-check.r to avoid rule engine cutoff

claudisa-uu:
This refactor shortens the package-check.r script to avoid length issues caused by the rule engine.


lwesterhof (approved):
LGTM!

PR #614, opened 2025-08-28
YDA-6514 - Moved the provenance action from pre status transition to post status transition for canceled publications.

kaur16:
YDA-6514 - Moved the provenance action from pre status transition to post status transition for canceled publications.


lwesterhof (approved):
LGTM!

PR #615, opened 2025-08-29
YDA-6525: expand publication exception logging

stsnel:
Change error handling code in the publication module to always log exception traces. These are often necessary for determining the cause of unexpected problems.


claudisa-uu (approved):
Looks good and works properly!

PR #616, opened 2025-09-01
YDA-6524: skip unsupported packages in pub.update

stsnel:
When updating publication endpoints, skip data packages with metadata schemas that are known to be unsupported. Trying to generate publication data (e.g. landing pages) for such data packages does not work and results in hard-to-read log clutter, so it's better to just skip these packages with an informational message.

### DIFF
--- a/publication.py
+++ b/publication.py
@@ -1535,6 +1535,29 @@ def update_publication(ctx: rule.Context,
         log.write(ctx, "update_publication: Not processing vault package, because initial status is " + status)
         return status
 
+    # Abort if data packages has a known unsupported metadata schema
+    try:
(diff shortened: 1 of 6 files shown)


claudisa-uu (commented):
Looks good and works properly. The only thing I see that could be improved is that on the client side no error or notification is returned.

The rule just prints

> `[publication] /tempZone/home/vault-core-0/test: OK`

(OK being the publication status) so anyone running the script without looking closely at the iRODS log might not notice that the vault package was not processed.

stsnel:
Thanks, the PR has been updated so that it puts the data package in an unrecoverable publication status when it has an unsupported metadata schema.

claudisa-uu (approved):
New changes tested and looking good!

PR #617, opened 2025-09-01
YDA-6527: skip faulty DPs in retention notif. job

stsnel:
In the data manager notification job for retention time of archived data packages, add error handling code for data packages without metadata.

Such data packages are invalid, but may be present on the system due to other problems. These data packages need to be ignored by the job. Log an informational message about this for technical admins.


claudisa-uu (approved):
Looks good to me!

PR #618, opened 2025-09-01
YDA-6528: improve DM group error handling

stsnel:
Add error handling code for situations where we are unable to determine the datamanager group of a research group, e.g. because it has no category.

Research groups without a category are invalid, but they can be created under various error conditions, so policies and reporting jobs need to be able to handle such cases.


folder.py
```diff
@@ -462,10 +462,15 @@ def folder_secure_fail(ctx: rule.Context, coll: str) -> None:
 def send_folder_secure_notification(ctx: rule.Context, coll: str, message: str) -> None:
     """Send notification about folder secure to relevant datamanagers"""
     if datamanager_exists(ctx, coll):
-        datamanagers = get_datamanagers(ctx, coll)
+        try:
+            datamanagers = get_datamanagers(ctx, coll)
+        except ValueError as e:
+            log.write(ctx, f"Unable to send folder secure notifications for <{coll}>: cannot get data managers: {str(e)}")
```
FuHsinyu, line 468: If the category is None in get_datamanagers(ctx, coll), the valueError and the log.write wont be triggered, is this expected? 
I created a case that a group has a removed category attrbute, and trigger the get_datamanagers function. The result shows "no log is reported", but an empty value [] for datamanagers 
And this is done through a publication process, however, no notification / error message is reported and the publication failed. As a result, no packages are copied to vault. I suppose it failed because no datamanager to send notification

folder.py
```diff
@@ -765,19 +770,47 @@ def get_status(ctx: rule.Context, path: str, org_metadata: List[Tuple[str, str]]
...
+       """
     group_name = collection_group_name(ctx, coll)
+    if group_name == "":
+        raise ValueError(f"Error: cannot determine group of collection {coll}.")
     category = group.get_category(ctx, group_name)
-
+    if category is None:
+        # Group without category is invalid, but can be created in various
+        # error conditions. In this case, it has no data manager.
+        return False
```
FuHsinyu: Same here, if the category is NOne, only False is returned but no error handling triggered

folder.py
stsnel: Thanks, you're right. The issue has been fixed by raising a `ValueError` in `get_datamanagers` and `datamanager_exists` if the group has no category.

folder.py
```diff
@@ -462,10 +462,15 @@ def folder_secure_fail(ctx: rule.Context, coll: str) -> None:
 def send_folder_secure_notification(ctx: rule.Context, coll: str, message: str) -> None:
     """Send notification about folder secure to relevant datamanagers"""
     if datamanager_exists(ctx, coll):
```
FuHsinyu, line 464: A minor suggestion: could this also be in try-except block? since datamanager_exists here is not covered by error handling

FuHsinyu:
LGTM!

PR #621, opened 2025-09-04
YDA-6534: update iCommands docs link

stsnel:
Update link to iCommands documentation on data transfer page, so that it links to the iCommands version that matches the iRODS server version. Yoda should be used with an iCommands version that matches the iRODS version on the server.


FuHsinyu (commented):
LGTM!

PR #625, opened 2025-09-08
YDA-6538 - Removed GROUP aggregate function from deprovision users script

kaur16:
YDA-6538 - Removed GROUP aggregate function from deprovision users script


claravox (approved):
LGTM!

PR #627, opened 2025-09-10
yda-6136-improve-copy-to-research-codes_3

FuHsinyu:
1. Improved copy-to-research flow by replacing legacy irods rule (iiCopyFolderToResearch) with irsync. This prevents potential issue of memory leakage. 
2. Piloted using delayExec for retry mechanism.
3. Added sending notification to user when copy-to-research successes or fails.  
Associated Portal PR: https://github.com/UtrechtUniversity/yoda-portal/pull/437 and https://github.com/UtrechtUniversity/yoda/pull/643/files

### DIFF
--- a/unit-tests/test_vault.py
+++ b/unit-tests/test_vault.py
@@ -8,22 +8,26 @@ from unittest import TestCase
 
 sys.path.append('..')
 
-from vault_utils import get_copy_folder_to_vault_irsync_command, get_sanity_checks_results_copy_to_vault_paths
+from vault_utils import get_copy_irsync_command, get_sanity_checks_results_copy_to_research_paths, get_sanity_checks_results_copy_to_vault_paths
 
 
 class VaultTest(TestCase):
 
-    def test_get_copy_folder_to_vault_irsync_command_with_vault_resc(self):
-        output = get_copy_folder_to_vault_irsync_command("/zoneName/home/research-foo/abc", "/zoneName/home/vault-foo/abc", "vaultResc", True)
+    def test_get_copy_irsync_command_with_vault_resc(self):
+        output = get_copy_irsync_command("/zoneName/home/research-foo/abc", "/zoneName/home/vault-foo/abc/original", "vaultResc", True)
         self.assertEqual(output, ["irsync", "-rK", "-R", "vaultResc", "i:/zoneName/home/research-foo/abc/", "i:/zoneName/home/vault-foo/abc/original"])
 
-    def test_get_copy_folder_to_vault_irsync_command_without_vault_resc(self):
-        output = get_copy_folder_to_vault_irsync_command("/zoneName/home/research-foo/abc", "/zoneName/home/vault-foo/abc", None, True)
+    def test_get_copy_irsync_command_without_vault_resc(self):
+        output = get_copy_irsync_command("/zoneName/home/research-foo/abc", "/zoneName/home/vault-foo/abc/original", None, True)
         self.assertEqual(output, ["irsync", "-rK", "i:/zoneName/home/research-foo/abc/", "i:/zoneName/home/vault-foo/abc/original"])
 
-    def test_get_copy_folder_to_vault_irsync_command_no_multithreading(self):
-        output = get_copy_folder_to_vault_irsync_command("/zoneName/home/research-foo/abc", "/zoneName/home/vault-foo/abc", "vaultResc", False)
+    def test_get_copy_irsync_command_no_multithreading(self):
+        output = get_copy_irsync_command("/zoneName/home/research-foo/abc", "/zoneName/home/vault-foo/abc/original", "vaultResc", False)
         self.assertEqual(output, ["irsync", "-rK", "-R", "vaultResc", "-N", "0", "i:/zoneName/home/research-foo/abc/", "i:/zoneName/home/vault-foo/abc/original"])
+
+    def test_get_copy_irsync_command_copy_from_research_mode(self):
+        output = get_copy_irsync_command("/zoneName/home/vault-foo/abc", "/zoneName/home/research-foo/abc", None, False)
+        self.assertEqual(output, ["irsync", "-rK", "-N", "0", "i:/zoneName/home/vault-foo/abc/", "i:/zoneName/home/research-foo/abc"])
 
     def test_get_sanity_check_results_copy_to_vault_paths_ok(self):
         output = get_sanity_checks_results_copy_to_vault_paths("/tempZone/home/research-foo", "/tempZone/home/vault-foo")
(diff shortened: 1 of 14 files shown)


vault.py
```diff
@@ -174,73 +176,264 @@ def api_vault_republish(ctx: rule.Context, coll: str) -> api.Result:
...
     if not collection.exists(ctx, coll_target):
-        return api.Error('TargetPathNotExists', 'The target you specified does not exist')
+        return api.Error('TargetPathNotExists',
+                         'The specified target does not exist')
 
-    # Check if user has READ ACCESS to specific vault package in collection coll_origin.
+    # Verify sub_coll_target package NOT already exists
+    if collection.exists(ctx, sub_coll_target):
+        return api.Error('PackageAlreadyPresentInTarget',
+                         'This datapackage is already present at the specified location')
```
claravox: It would be good if we could temporarily disable the "copy package to research area" button (in the portal) before the package is being copied notification is loaded. Otherwise I could imagine users double clicking the button thinking that nothing is happening.

FuHsinyu: Agreed, looks like a better solution visually. Fyi, the quoted part is not for preventing double clicking, but to check if the package was copied previously. The part for preventing double clicking is in other places (line 267 of vault.py. So if user double clicks, only the first time will be converted to copy-to-research. Second click will be ignored

claravox:
So can we remove iiCopyFolderToResearch then? Or not because I see that deposit still uses it?

FuHsinyu:
> So can we remove iiCopyFolderToResearch then? Or not because I see that deposit still uses it?

I am not sure either. I also notice it is used elsewhere. I can create a ticket for that.

unit-tests/test_vault.py
```diff
@@ -8,23 +8,31 @@
...
+        output = get_copy_folder_irsync_command("/zoneName/home/research-foo/abc", "/zoneName/home/vault-foo/abc", None, True)
         self.assertEqual(output, ["irsync", "-rK", "i:/zoneName/home/research-foo/abc/", "i:/zoneName/home/vault-foo/abc/original"])
 
-    def test_get_copy_folder_to_vault_irsync_command_no_multithreading(self):
-        output = get_copy_folder_to_vault_irsync_command("/zoneName/home/research-foo/abc", "/zoneName/home/vault-foo/abc", "vaultResc", False)
+    def test_get_copy_folder_irsync_command_no_multithreading(self):
+        output = get_copy_folder_irsync_command("/zoneName/home/research-foo/abc", "/zoneName/home/vault-foo/abc", "vaultResc", False)
         self.assertEqual(output, ["irsync", "-rK", "-R", "vaultResc", "-N", "0", "i:/zoneName/home/research-foo/abc/", "i:/zoneName/home/vault-foo/abc/original"])
 
+    def test_get_copy_folder_irsync_command_with_research_mode(self):
```
claravox: Could we make this test look more like how a copy to research command would look like?

vault.py
```diff
@@ -174,73 +176,259 @@ def api_vault_republish(ctx: rule.Context, coll: str) -> api.Result:
...
+    )
+    irsync_success = _execute_command(ctx, irsync_command, "irsync copy")
+
+    # Restore original permission
+    permission_restored = True
+    if rods_permission != "own":
+        restore_cmd = ["ichmod", "-M", rods_permission, rodsadmin_fullname, parent_coll]
+        permission_restored = _execute_command(ctx, restore_cmd, "Permission restoration")
+
+    if not irsync_success or not permission_restored:
```
claravox: Shouldn't we check after this first line for if the irsync was a success before changing the permissions?

FuHsinyu: This restores the original permission of rods to the **parent folder**. so regardless irsync's result, the permission should be restored. Otherwise it added an extra permission to target_coll's parent folder.

vault.py
```diff
@@ -174,73 +176,259 @@ def api_vault_republish(ctx: rule.Context, coll: str) -> api.Result:
...
+            return True
+        log.write(ctx, f"{action_name} failed: {' '.join(command)} (code {returncode})")
+        return False
+    except Exception as e:
+        log.write(ctx, f"Error during {action_name}: {str(e)}")
+        return False
+
+
+def handle_retry_operation(ctx: rule.Context, coll_origin: str, coll_target: str, receiver: str, retry_count: int) -> bool:
+    """Manage retry logic with delays"""
```
claravox: Can you add to this comment/func to make it clear that it is for copy to research?

vault.py
```diff
@@ -174,73 +176,268 @@ def api_vault_republish(ctx: rule.Context, coll: str) -> api.Result:
...
+
+    :param ctx:         Combined type of a callback and rei struct
+    :param coll_origin: Origin data collection in vault space
+    :param coll_target: Target collection in research space
+    :param receiver:    User to notify of success/failure
+    :param retry_str:   Current retry attempt as string
+
+    :returns:           True if operation succeeded or entered retry logic, False if target already existed
+    """
+    log.write(ctx, f"Starting vault copy to research: {coll_origin} -> {coll_target}, attemp #{retry_str}")
```
claravox:
```suggestion
    log.write(ctx, f"Starting vault copy to research: {coll_origin} -> {coll_target}, attempt #{retry_str}")
```

vault.py
```diff
@@ -174,73 +176,268 @@ def api_vault_republish(ctx: rule.Context, coll: str) -> api.Result:
...
+    # Concatenate coll_target with package name to create sub_coll_target
+    package_target_name = coll_origin.rstrip('/').split('/')[-1]
+    sub_coll_target = f"{coll_target.rstrip('/')}/{package_target_name}"
 
-    # Now check whether target collection already exist.
-    if collection.exists(ctx, new_package_collection):
-        return api.Error('PackageAlreadyPresentInTarget', 'This datapackage is already present at the specified place')
+    # Concatenate coll_target with package name to create sub_coll_target
+    sub_coll_target = "{}/{}".format(coll_target.rstrip('/'), package_target_name)
+    log.write(ctx, "Copyinga data package from vault to research: {}".format(sub_coll_target))
```
claravox:
```suggestion
    log.write(ctx, "Copying data package from vault to research: {}".format(sub_coll_target))
```

vault.py
```diff
@@ -174,73 +176,228 @@ def api_vault_republish(ctx: rule.Context, coll: str) -> api.Result:
...
-    :param coll_origin: Collection of data package to copy
-    :param coll_target: Collection to copy data package to
+    :param coll_origin: Origin data collection in vault space
+    :param coll_target: Target collection path in research space
 
     :returns: API status
     """
+    # Validate target location
     zone = user.zone(ctx)
+    if coll_target == f"/{zone}/home":
```
lwesterhof: Can we rewrite these coll_target checks using [pathutil.info](https://github.com/UtrechtUniversity/yoda-ruleset/blob/development/util/pathutil.py#L74)?

vault.py
```diff
@@ -314,11 +314,13 @@ def execute_irsync_copy(ctx: rule.Context, coll_origin: str, coll_target: str) -
     rodsadmin_fullname = f"rods#{zone}"
     parent_coll = os.path.dirname(os.path.normpath(coll_target))
 
+    if user.user_type(ctx) != 'rodsadmin':
```
lwesterhof: We can use user.is_admin(ctx) instead.

vault.py
```diff
@@ -314,11 +314,13 @@ def execute_irsync_copy(ctx: rule.Context, coll_origin: str, coll_target: str) -
...
+        log.write(ctx, "Error: insufficient permissions to run ARB data update rule.")
```
lwesterhof: Copy paste error

lwesterhof (changes requested):

vault.py
```diff
@@ -174,73 +176,230 @@ def api_vault_republish(ctx: rule.Context, coll: str) -> api.Result:
...
-    if coll_target == "/" + zone + "/home":
-        return api.Error('HomeCollectionNotAllowed', 'Please select a specific research folder for your datapackage')
-
-    # Check if target is a research folder. I.e. none-vault folder.
+    # Check target is not in vault space
     parts = coll_target.split('/')
-    group_name = parts[3]
-    if group_name.startswith('vault-'):
-        return api.Error('RequiredIsResearchArea', 'Please select a specific research folder for your datapackage')
+    if len(parts) > 3 and parts[3].startswith('vault-'):
```
lwesterhof: Can we rewrite this condition using [pathutil.info](https://github.com/UtrechtUniversity/yoda-ruleset/blob/development/util/pathutil.py#L74)?

vault.py
FuHsinyu: I added, but not sure if my fix is proper?

vault.py
```diff
@@ -174,73 +176,223 @@ def api_vault_republish(ctx: rule.Context, coll: str) -> api.Result:
...
+        str(next_attempt),
+        str(wait_seconds)
+    ]
+    if not _execute_command(ctx, command, f"copy-to-research (retry #{next_attempt} for {coll_origin} -> {coll_target})"):
+        return False
+
+    return True
+
+
+def _execute_command(ctx: rule.Context, command: list[str], action_name: str = "") -> bool:
```
lwesterhof: We want to use msiExecCmd instead of subprocess. This function should never exist without validating the command.

vault.py
lwesterhof: Can we simplify it by combining it with the condition on l196?

lwesterhof (changes requested):

PR #628, opened 2025-09-12
YDA-6554: fix DP retention time notifications

stsnel:
Fix notifications job for exceeding the retention time of archived data packages. The function now processes only top-level collections of archived data packages. This resolves issues where the job will send notifications to users for other collections (e.g. for archived data packages that are already in the trash).

### DIFF
--- a/util/pathutil.py
+++ b/util/pathutil.py
@@ -129,6 +129,6 @@ def is_archived_datapackage_path(path: str) -> bool:
 
     :param path: Path to check
 
-    :returns: True if path refers to top-leve collection of data package. Else False.
+    :returns: True if path refers to top-level collection of data package. Else False.
     """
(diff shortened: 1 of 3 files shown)


util/pathutil.py
```diff
@@ -122,3 +122,13 @@ def test(r: str, space: Space) -> Tuple[Space, str, str, str] | None:
...
             or test('^/([^/]+)()(?:/(.+))?$',                         Space.OTHER)
             or (Space.OTHER, '', '', '')))  # (matches '/' and empty paths)
+
+
+def is_archived_datapackage_path(path: str) -> bool:
+    """Tests if a path is the top-level collection of an archived data package
+
+    :param path: Path to check
+
+    :returns: True if path refers to top-leve collection of data package. Else False.
```
claravox:
```suggestion
    :returns: True if path refers to top-level collection of data package. Else False.
```

claravox (approved):
LGTM!

PR #630, opened 2025-09-16
YDA-6556: fix sort category list in group manager

leonidastri:
This fixes the issue of categories not being sorted alphabetically in group manager module. It also adds api test to test if categories, subcategories and groups are sorted alphabetically.


stsnel (approved):
Looks good to me, also great work on the thorough API test :+1: 

I've split off a separate low-priority ticket for the issue that occurs when portal caching is enabled (YDA-6570)

PR #632, opened 2025-09-17
YDA-6549 fix for bug with dismissing notifications

stsnel:
This fixes a bug where rodsadmin users were unable to dismiss notifications in the portal. This was caused by a policy change in YDA-5813 which removed policy logic that allowed rodsadmin users to remove any user metadata.

This fix restores the ability of rodsadmin users to perform any metadata changes that regular users can perform, which also fixes the problem with dismissing notifications.


FuHsinyu:
LGTM!

PR #633, opened 2025-09-18
YDA-6567 - Update of metadata on landingpage fails with new version of datapackage

kaur16:
After the vault ingest, the update publication was not running. The metadata of the data package (second version) had all the required AVUs to start the update publication action. However, the first version of the data package was missing baseDOIMinted AVU which resulted in error (line 801). After minting base DOI (line 1056), added a condition to see if the data package is a new version of a previous publication. If yes, then setting the baseDOIMinted key from previous publication as the latest one and save the publication state.


lwesterhof (approved):
LGTM!

PR #634, opened 2025-09-18
YDA-6573: fix data package submit policy

stsnel:
Fix issue where data package in group with no data managers was not automatically accepted due to a
regression in the policy changes in YDA-6528.

Also extract policy logic to separate function and add test coverage to reduce risks of related regressions.


FuHsinyu (approved):
LGTM! Clear logic and good tests with efficient test coverage

PR #635, opened 2025-09-22
YDA-6563: fix license names mismatch

claudisa-uu:
This PR renames the "No Derivatives" license files to match the enum options of the metadata schemas. We thought it was easier to update the file names rather than the metadata schema enums, since in that case we would need to release new versions of the metadata schemas.

NOTE: When the playbook is rerun after this change, the old files will still be there and will need to be cleaned up.


lwesterhof (approved):
LGTM!

PR #636, opened 2025-09-24
YDA-6580 - Modified transform-existing-publications to include latest changes in publication process.

kaur16:
YDA-6580 - Modified transform-existing-publications to include baseDOIMinted changes in the publication process.


lwesterhof (approved):
LGTM!

PR #640, opened 2025-10-22
YDA-6629: fix coll match deposit module overview

stsnel:
Fix matching of deposit group collections in the deposit module overview function. Checking whether the collection name contains "deposit-" is insufficient, since non-deposit collection names can also contain "deposit-". Use the default pathutil.info function instead, which handles this correctly.


lwesterhof (approved):
LGTM

PR #642, opened 2025-10-30
Yda 5988: move data to vault in archive process

FuHsinyu:
The PR adds an option to submit the data package and remove it once the move-to-vault action is successful
Link to Portal: https://github.com/UtrechtUniversity/yoda-portal/pull/440


folder.py
```diff
@@ -427,9 +439,11 @@ def folder_secure_succeed_avus(ctx: rule.Context, coll: str, group_name: str) ->
     # Remove target AVU on source folder. This should be done after all possibly failing steps
     # have occurred in folder_secure (any "return False" steps), so that if those trip a retry state,
     # on retry folder_secure can reuse the target from before.
-    if not group_name.startswith("deposit-") and constants.IICOPYPARAMSNAME in org_metadata:
-        if not avu.rmw_from_coll(ctx, coll, constants.IICOPYPARAMSNAME, "%", True):
-            return False
+    # Check the collection still exists, as research collection may be deleted after secured.
+    if collection.exists(ctx, coll):
+        if not group_name.startswith("deposit-") and constants.IICOPYPARAMSNAME in org_metadata:
```
FuHsinyu: Is it still needed to check if the group_name start with deposit? since if it is deleted collection.exists will not pass

policies_folder_status.py
```diff
@@ -234,3 +236,31 @@ def post_status_transition(ctx: rule.Context,
         submitter = folder.get_submitter(ctx, path)
         message = "Data package rejected for vault"
         notifications.set(ctx, actor, submitter, path, message)
+
+
+def delete_research_package(ctx: rule.Context, path: str) -> None:
```
FuHsinyu: Is this a proper place to place this function?

lwesterhof: We can combine it with these lines from deposit: https://github.com/UtrechtUniversity/yoda-ruleset/blob/61ad749d468754c20e0f7c34e02e912be8311f9d/policies_folder_status.py#L221-L225

Just needs a condition and then we do not need the new function.

policies_folder_status.py
```diff
@@ -234,3 +236,30 @@ def post_status_transition(ctx: rule.Context,
...
+
+    :param ctx:  Combined type of a callback and rei struct
+    :param path: Collection path in research zone
+    """
+    try:
+        # Get AVU to decide keep copy or delete research package
+        keep_package = avu.get_attr_val_of_coll(ctx, path, constants.KEEP_RESEARCH_COPY)
+
+        # Delete collection if KEEP_RESEARCH_COPY is "False"
+        if keep_package == "False":
```
lwesterhof: We can write this as `if not keep_package:`.

Might be more intuitive if we name the variable delete_research_copy, because that is the happy flow of this condition.

folder.py
lwesterhof: No, we can just use collection.exists, same for l301.

lwesterhof (changes requested):

PR #643, opened 2025-11-12
YDA-6664 - Update condition for updating existing groups in SRAM enabled instances

kaur16:
YDA-6664 - Update condition for updating existing groups in SRAM enabled instances


lwesterhof (approved):
LGTM

PR #645, opened 2025-11-18
YDA-6667: fix MSI_OPERATION_NOT_ALLOWED error when targeting consumer with iBridges

claudisa-uu:
When targeting consumer with iBridges, an MSI_OPERATION_NOT_ALLOWED error is raised when msiSetACL is attempted on replication and revision. This PR fixes the issue by moving that call in the replication and revision batch job itself instead of earlier.


lwesterhof (approved):
LGTM

PR #649, opened 2025-11-26
YDA-6703: add rightsIdentifier to DataCite metadata

lwesterhof:
https://support.datacite.org/docs/where-does-the-license-information-come-from-in-datacite-commons


claudisa-uu (approved):
Looks good and works properly!

PR #651, opened 2025-12-01
YDA-6711: format discipline in a way that DataCite understands

lwesterhof:


claudisa-uu (approved):
Looks good and works as expected!

PR #652, opened 2025-12-04
YDA-6636: add function to exempt specific resources from replication

claudisa-uu:
This PR adds a function that allows specific configurable resources to be exempt from replication policy


lwesterhof (approved):
LGTM

PR #656, opened 2025-12-09
YDA-6652: Disable zip download on landing pages if Anubis enabled

claravox:


claravox:
Likely not needed with the javascript change

claravox:
Not needed with the javascript change of the landing pages

PR #659, opened 2026-01-14
YDA-6746: add full publication date to DataCite metadata

lwesterhof:


json_datacite.py
```diff
@@ -322,14 +322,32 @@ def get_contributors(combi: Dict) -> List:
...
+    """Return list of dates in DataCite format.
+
+    :param combi: Combined JSON file that holds both user and system metadata
+
+    :returns: JSON element with dates in DataCite format
+    """
+
+    # Format dates for DataCite: https://datacite-metadata-schema.readthedocs.io/en/4.6/properties/date/
+    publication_date = combi.get('System', {}).get('Publication_Date')
+    publication_date = parser.parse(publication_date)
```
FuHsinyu, line 334: Would it be nice to have a defensive check in case publication_date parsed is None and then crash? e.g, add `if publication_date: `

lwesterhof: We know that it is there, we check that here: https://github.com/UtrechtUniversity/yoda-ruleset/blob/c1d955e8a2b174f791a513d9da8247bb0927ac7e/publication.py#L826-L830

PR #660, opened 2026-01-14
Yda 6626 make data request assignee mandatory

FuHsinyu:
Screenshot from 2026-01-14 15-35-18

The alert message looks like this

### DIFF
--- a/datarequest.py
+++ b/datarequest.py
@@ -1420,7 +1420,8 @@ def api_datarequest_assignment_submit(ctx: rule.Context, data: Dict, request_id:
         permitted_to_read = [GROUP_DM, GROUP_PM]
         if 'assign_to' in data:
             permitted_to_read = permitted_to_read + data['assign_to'][:]
-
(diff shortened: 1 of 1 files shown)


lwesterhof (commented):
This is a nice defensive measure, in addition we also want to fix the datarequest schema. So that there is a minimum of 1 selected DAC members enforced by the schema.

claudisa-uu (commented):
Looks and works good, however this change doesn't necessarily make the assignee field mandatory, it just catches the error that occurs when the form is submitted without the assignee. 

To make the field mandatory, we would need to change the schema itself, like Lazlo mentioned earlier. Since that is a more complex issue (I believe it would require a new schema version?), maybe we can still merge this but I would change the commit message to reflect the actual changes.

claudisa-uu (approved):
Looks good to me, I think we can merge even with the failing test since we established that that is unrelated to these changes.

PR #661, opened 2026-01-20
YDA-6752: fix empty checksum function 

claravox:


util/collection.py
claravox, line 1: subcollection function did not seem to be working properly. For example giving a path like this: `/tempZone/home/research-default-3/folder01//tempZone/home/research-default-3/folder01/subfolder01`. I think we should add a test but in the interest of time I added only a basic one

stsnel (approved):
Nice, LGTM! :+1:

PR #665, opened 2026-01-28
YDA-6624: set correct ACLs to data request files

claudisa-uu:
This PR ensures that the correct ACLs are set to all data request files so that system operations (such as email notifications) work properly.


lwesterhof (approved):
LGTM!

PR #666, opened 2026-02-02
YDA-6616: move parsing of open access link to combiJson creation

claudisa-uu:
This PR moves the URL parsing from the landing page creation to the combiJson creation, where the open_access_link is first generated. This solves the 422 HTTP response from DataCite, as the link was not parsed before sending it to DataCite.


lwesterhof (approved):
LGTM

PR #667, opened 2026-02-03
YDA-6780: add tool for updating schema single DP

stsnel:
Add tool for checking and updating schema of single data package in the vault. This is useful when troubleshooting data package metadata issues, as well as for schema upgrade scenarios that involve only a subset of data packages on a Yoda environment.


lwesterhof (approved):
LGTM!

PR #670, opened 2026-02-04
YDA-6205: decrease number of queries needed to export monthly storage…

lwesterhof:
… statistics

### DIFF
--- a/stats.py
+++ b/stats.py
@@ -1,7 +1,7 @@
 """Functions for statistics module."""
 from __future__ import annotations
 
-__copyright__ = 'Copyright (c) 2018-2025, Utrecht University'
+__copyright__ = 'Copyright (c) 2018-2026, Utrecht University'
 __license__   = 'GPLv3, see LICENSE'
 
 from datetime import datetime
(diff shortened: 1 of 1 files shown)


stats.py
```diff
@@ -276,25 +267,27 @@ def api_resource_monthly_category_stats(ctx: rule.Context) -> api.Result:
...
-        date_reference = "{}_{}".format(min_year, '%0*d' % (2, min_month))
+        date_reference = f"{min_year}_{min_month:02}"
         storage_dates.append(date_reference)
 
         for category in categories:
             # for all groups in category
-            groups = get_groups_on_categories(ctx, [category])
+            groups = groups_cache[category]
             for group in groups:
                 if group.startswith(('research', 'deposit', 'intake', 'grp')):
```
FuHsinyu, line 292: A possible minor improvement: I am thinking if we could also cache this to e.g  `eligible_groups`, so that `groups.startswith` wont be used during the loop for every group, especially for a large number of groups

stats.py
FuHsinyu, line 227: I have only reviewed this function due to time constraint, and it looks good and significant improvement! The improvement scales with the earliest month of the group creation, which in my test test case it wasn't much as I created all groups this month. 

Test case: 4000 groups, created this month, reduce time from 48sec - 43 sec after the PR

claudisa-uu (approved):
Looks good to me! I tested the changes on 4000 groups with simulated storage metadata across all 2025, they seem to save a couple of seconds per loop (compared to development).

PR #671, opened 2026-02-04
YDA-6781: improve schema validation messages

stsnel:
Ensure that schema validation human-readable messages include specific information in case of multiple undefined attributes or in case of a non-standard error message.

As a part of this change, the validation functions were refactored so that these functions could be covered by unit tests.


lwesterhof (approved):
LGTM!

PR #672, opened 2026-02-04
YDA-6777: add baseDOIMinted as expected AVU for first version of publication

claudisa-uu:
This PR adds the baseDOIMinted to the list of expected AVUs for the first version of a publication. This also works for the troubleshooting tool.


stsnel (approved):
LGTM 👍

PR #674, opened 2026-02-13
YDA-6803: fix provider/consumer detection ARB

stsnel:
The ARB script failed with an exception on some consumer environments because it assumed server configs always have an icat_host key. This fix ensures the script also works correctly if that is not the case.


leonidastri (approved):
LGTM!

PR #675, opened 2026-02-13
YDA-6755: fix ARB script for missing vault paths

stsnel:
Make the ARB update script robust against resources with vault paths that do not exist, by not processing them (but printing a notification message instead). The bundleResc resource is now ignored silently, since it is not used for regular storage.


leonidastri (approved):
LGTM!

PR #685, opened 2026-03-12
YDA-6793: add timeouts ARB update script

stsnel:
Add a timeout to the ARB update script. ARB update jobs are usually executed very minute. This change should prevent
ARB update jobs from piling up if the environment has many storage resources and is responding very slowly (e.g. due to high load).


lwesterhof (approved):
LGTM!

PR #687, opened 2026-03-13
YDA-6869: adjust auth PEPs for iRODS 5

stsnel:
iRODS 5 apparently uses two different PEPs for authentication: one for the anonymous account and for the other accounts. This ensures that we handle both of them correctly.

This change needs a full iRODS service restart in order to test it.


lwesterhof (approved):
LGTM

PR #691, opened 2026-03-18
Add integration tests for stats/get user groups

stsnel:
Add integration tests for the stats.get_user_groups_for_stats function to be better able to detect future regressions there.


lwesterhof (approved):
LGTM

PR #695, opened 2026-03-20
YDA-6901: fix category statistics without coll

stsnel:
Ensure that the storage statistics job keeps working if it encounters groups without any available storage data. This can happen if we have any loose groups without a group collection (e.g. group research-abc exists, but collection /zone/home/research-abc does not).

Although this is not an expected situation, it can happen if an error occurred when removing a group, or if a rodsadmin tried to manually remove data without cleaning up the group.


lwesterhof (approved):
LGTM

PR #699, opened 2026-03-27
Fix: improve (sub)category names' validation

FuHsinyu:
Improved the functions for (sub)cat name validation and added unit tests.

The improvements are based on the design doc: https://utrechtuniversity.github.io/yoda/design/conventions/#category-names 

Also most of them are tested through portal's feature of adding group and cat/subcat,  to ensure the design doc correspond to portal's tech design


lwesterhof:
Can you also add a unit test to cover this? And do note that we have to change is_valid_subcategory as well, since there uppercase letters are allowed.

PR #703, opened 2026-04-01
YDA-6175: only store subset of datarequest on AVUs

lwesterhof:


stsnel (approved):
LGTM! 👍 

I've split off a separate ticket for removing the now unused AVUs (YDA-6927). 

I've performed limited testing with larger input sizes, due to the size restrictions we have because of issue YDA-6004. I'll add a note to that ticket to specifically retest this after that ticket has been resolved as well.

PR #704, opened 2026-04-01
YDA-6916: add source path validation for archiving script

FuHsinyu:
Added a check for trailing slash. A trailing slash can result in issue in msiArchiveCreat.cc not be able to find the right source directory in archive mode.


kaur16 (approved):
Looks good to me!

PR #705, opened 2026-04-02
Deprecate legacy iiRemoveAVUs function

lwesterhof:


stsnel (approved):
The changes in the ruleset looks good to me👍 

Please consider changing the commit message to say that the function has been removed rather than deprecated.

PR #709, opened 2026-04-08
Add constants for vault archive statuses

lwesterhof:


vault_archive.py
```diff
@@ -262,22 +268,22 @@ def vault_extract_archive(ctx: rule.Context, coll: str) -> str:
 
 def update(ctx: rule.Context, coll: str, attr: str | None) -> None:
     if pathutil.info(coll).space == pathutil.Space.VAULT and attr not in (constants.IIARCHIVEATTRNAME, constants.UUPROVENANCELOG) and vault_archival_status(ctx, coll) == "archived":
```
stsnel: I think, to be consistent, "archived" could also be replaced with a constant here?

vault_archive.py
stsnel: I think, to be consistent, "archived" could also be replaced with a constant here?

stsnel: I think, to be consistent, "archive" could also be replaced with a constant here?

stsnel: I think, to be consistent, "extract" could also be replaced with a constant here?

util/bagit.py
```diff
@@ -44,16 +43,6 @@ def manifest(ctx: rule.Context, coll: str) -> str:
     ]) + "\n"
 
 
-def status(ctx: rule.Context, coll: str) -> str | bool:
```
stsnel, line 47: If this function is removed here, I believe the references to this function in `vault_download.py` need to be updated.

stsnel (commented):
The change looks good to me in general 👍 

Please find some questions/remarks about details below.

stsnel (approved):
LGTM :+1: !

PR #717, opened 2026-04-17
Util.AVU: fix wildcard match function

stsnel:
Fix wildcard match function for AVUs so that it supports wildcards in any position.


lwesterhof (approved):
LGTM!

PR #721, opened 2026-04-30
YDA-6396: add new epos-msl-1 metadata schema

lwesterhof:


claudisa-uu:
The dropdown menus are not ordered alphabetically; it's technically not essential as you can type in the field to search, but maybe it could be a point of improvement?

claudisa-uu (changes requested):
If I fill in metadata in the epos-msl-0 schema and copy the file over to the epos-msl-1 group, I get an error saying that the metadata file is not compliant with the schema. There should be an option to transform the metadata into the new schema.

claudisa-uu:
There are also issues with required fields: I have filled in every field that looks required (with asterisks or highlighted in red), but the "Required for Vault" bar is not filled in and I cannot submit for vault.

claudisa-uu:
> If I fill in metadata in the epos-msl-0 schema and copy the file over to the epos-msl-1 group, I get an error saying that the metadata file is not compliant with the schema. There should be an option to transform the metadata into the new schema.

I now correctly get the prompt to transform metadata, but after transformation the system still says that metadata is not compliant.

PR #722, opened 2026-04-30
YDA-6961: log spOption value for new connections

stsnel:
This is useful for technical admins when trying to find log messages related to relevant sessions in the context of troubleshooting client problems. For example, when a user who is experiencing a technical issue is connecting to iRODS using multiple clients/interfaces simultaneously.


lwesterhof (approved):
LGTM!

PR #724, opened 2026-05-07
YDA-6927: add script to clean up obsolete AVUs from data request objects

claudisa-uu:
This PR adds a tool script to clean up obsolete AVUs from data request objects

### DIFF
--- a/tools/clean-up-datarequest-avus.r
+++ b/tools/clean-up-datarequest-avus.r
@@ -25,26 +25,29 @@ def remove_obsolete_avus(ctx, zone, dryrun):
                            "COLL_NAME, DATA_NAME, META_DATA_ATTR_NAME, META_DATA_ATTR_VALUE",
                            f"COLL_NAME like '/{zone}/home/datarequests%' AND DATA_NAME like '%datarequest%'")
 
-    if len(list(query)) > 0:
-        count = 0
-        for row in query:
-            coll, obj, attr, value = row[0], row[1], row[2], row[3]
-            if attr not in VALID_AVUS:
-                ctx.writeLine("stdout", f"Obsolete AVU found for object {coll}/{obj}: attribute '{attr}' with value '{value}'")
-                count += 1
+    count = 0
+    found_any = False
+    for row in query:
+        found_any = True
+        coll, obj, attr, value = row[0], row[1], row[2], row[3]
+        if attr not in VALID_AVUS:
+            ctx.writeLine("stdout", f"Obsolete AVU found for object {coll}/{obj}: attribute '{attr}' with value '{value}'")
+            count += 1
 
-                if not dryrun:
-                    ctx.writeLine("stdout", "Deleting...")
-                    msi.mod_avu_metadata(ctx, "-d", f"{coll}/{obj}", "rm", attr, value, "")
+            if not dryrun:
+                ctx.writeLine("stdout", "Deleting...")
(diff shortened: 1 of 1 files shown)


tools/clean-up-datarequest-avus.r
```diff
@@ -0,0 +1,90 @@
...
+
+# Get all datarequest data objects in the current zone
+def get_datarequest_objs(ctx, zone):
+    iter = genquery.Query(ctx,
+                          "COLL_NAME, DATA_NAME",
+                          f"COLL_NAME like '/{zone}/home/datarequest%'")
+
+    datarequest_objs = []
+    for row in iter:
+        if row[1] == "datarequest.json":
```
leonidastri: Suggestion: Maybe rename iter variable, so it does not shadow Python built-in iter()

tools/clean-up-datarequest-avus.r
```diff
@@ -0,0 +1,90 @@
...
+import genquery
+
+import msi
+
+VALID_AVUS = ['assignedForReview', 'endOfReviewPeriod', 'owner', 'reviewedBy', 'status', 'title']  # List of valid datarequest AVUs
+
+
+# Get all datarequest data objects in the current zone
+def get_datarequest_objs(ctx, zone):
+    iter = genquery.Query(ctx,
```
leonidastri: Suggestion: Maybe rename iter variable, so it does not shadow Python built-in iter()

leonidastri: For discussion: Could we combine these two genQueries into a single query that retrieves both the data objects and their AVUs in one pass? Right now we do one query to list objects and then an additional query per object to fetch metadata, which may become expensive on larger zones.

That said, the current split approach is quite readable, so it may not be worth changing unless this is expected to run against a large number of data request objects.

leonidastri (commented):
LGTM! The code is readable and easy to understand. I only have a few minor suggestions, though it depends on whether changing them is worthwhile, as some alternatives may make the code more complex and less readable.

tools/clean-up-datarequest-avus.r
claudisa-uu: I thought about making a single function to both fetch objects and delete AVUs when I was first implementing the script, but I found it more readable to keep them separate. It is true however that, with the current approach, the AVU query is called for every datarequest object that is found, so it might become expensive for larger environments. I have combined the queries into a single one for better efficiency, hopefully it's still readable.

tools/clean-up-datarequest-avus.r
```diff
@@ -0,0 +1,83 @@
...
+VALID_AVUS = ['assignedForReview', 'endOfReviewPeriod', 'owner', 'reviewedBy', 'status', 'title']  # List of valid datarequest AVUs
+
+
+# Remove all obsolete datarequest AVUs in the current zone
+def remove_obsolete_avus(ctx, zone, dryrun):
+    query = genquery.Query(ctx,
+                           "COLL_NAME, DATA_NAME, META_DATA_ATTR_NAME, META_DATA_ATTR_VALUE",
+                           f"COLL_NAME like '/{zone}/home/datarequests%' AND DATA_NAME like '%datarequest%'")
+
+    if len(list(query)) > 0:
```
leonidastri: I think we can remove this check

leonidastri:
LGTM!

PR #727, opened 2026-05-12
yda-external-user-sync- Replace is_admin with is_rodsadmin in external users sync for SRAM

kaur16:
yda-external-user-sync- Replace is_admin with is_rodsadmin in external users sync for SRAM


lwesterhof (approved):
LGTM

PR #733, opened 2026-05-18
YDA-6971: remove size column deposit overview

stsnel:
Calculating collection size synchronously takes too much time when there are large collection trees in pending deposits.


lwesterhof (approved):
LGTM

PR #735, opened 2026-05-19
YDA-6572: add workaround for correct total number of results in search page

claudisa-uu:
This PR adds a new query without offset or limit to get the correct total number of results from the search, as it seems that GenQuery's total_rows() function causes issues when a non-zero offset is provided.


lwesterhof:
Is probably fixed with https://github.com/UtrechtUniversity/yoda/pull/737

PR #739, opened 2026-05-26
YDA-6977: fix get user group function stats

stsnel:
This ensures that we don't exceed the maximum GenQuery size
when a non-rodsadmin user is viewing group statistics.


lwesterhof (approved):
LGTM!

PR #745, opened 2026-06-11
sram-sync-fixes - Fixed the SRAM sync and external user sync script.

kaur16:
sram-sync-fixes - Fixed the SRAM sync and external user sync script.


lwesterhof (approved):
LGTM

PR #748, opened 2026-06-12
YDA-7004: refactor util.config

stsnel:
Refactor the configuration code to accomplish the following:
- Nearly all code has been moved to the Configuration object, so that it is testable.
- Added unit tests for the key functions
- Deferred any exceptions that occurred during parsing the configuration file until we are past the compilation stage and authentication PEPs. At that time, we have proper error logging, and we can tell the technical administrator what's going on, rather than failing in a way that only shows a generic rule engine error message.

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -59,9 +59,9 @@ from vault                    import *
 from vault_deaccession        import *
 from epic                     import *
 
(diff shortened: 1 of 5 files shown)


lwesterhof (approved):
Nice, LGTM!

PR #751, opened 2026-06-15
YDA-7048: fix import issue after config refactor

stsnel:
In YDA-7004, the config module was refactored. The change in __init__.py introduced an import ordering issue that only appeared when the data package archiving module was enabled (because of the order of transitive imports). This fix ensures that the code in vault_archive.py also imports the correct config variable .


lwesterhof (approved):
LGTM

PR #752, opened 2026-06-16
YDA-7037: add WebDAV test suite

stsnel:


leonidastri (commented):
The scenarios provide a good coverage of the webdav interface. The code is clean and easy to understand. Maybe we could add a scenario for uploading/handling files which contain single quotes or spaces in the filename.

In general, LGTM!

stsnel:
> The scenarios provide a good coverage of the webdav interface. The code is clean and easy to understand. Maybe we could add a scenario for uploading/handling files which contain single quotes or spaces in the filename.
> 
> In general, LGTM!

Thanks :+1: , I've now added test cases for handling names with quotes and spaces.

PR #753, opened 2026-06-18
YDA-7061: deduplicate agent started log msgs

stsnel:
Print agent started log messages only once per session, even if pre-authentication PEPs are triggered multiple times. This also migrates the pre-authentication PEP functions to Python.


lwesterhof (approved):
LGTM!

PR #756, opened 2026-06-29
YDA-7031

lwesterhof:
Needs https://github.com/UtrechtUniversity/yoda-portal/pull/519

### DIFF
--- a/unit-tests/test_util_api.py
+++ b/unit-tests/test_util_api.py
@@ -3,12 +3,28 @@
 __copyright__ = 'Copyright (c) 2023-2026, Utrecht University'
 __license__   = 'GPLv3, see LICENSE'
 
+import base64
 import sys
+import zlib
+from typing import Optional, Union
 from unittest import TestCase
 
 sys.path.append('../util')
 
-from api import _check_type
+import jsonutil
+from api import _api, _check_type
+
+
+# --- Test helpers for encoding/decoding API input ---------------------------
+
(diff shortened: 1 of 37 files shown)


stsnel (approved):
LGTM! 👍 

Since I've added some code to the PR, I think it would be good to have an independent second review for this PR, with a focus on the commits that have been added later.

claravox (changes requested):
Thanks for adding these unit tests, they help a lot!
Generally looks good, I am wondering if it would be helpful to use assertDictEqual in cases where we compare dictionaries (see https://stackoverflow.com/questions/34414326/why-is-assertdictequal-needed-if-dicts-can-be-compared-by)
Also there is a merge conflict?

unit-tests/test_util_api.py
```diff
@@ -0,0 +1,182 @@
...
+    def test_api_wrapper_skips_type_check_for_untyped_argument(self):
+        # Arguments without a type hint are not type-checked at all.
+        def f(ctx, anything):
+            return anything
+
+        wrapped = _api(f)
+        result = wrapped(DummyContext(), encode_input({'anything': {'nested': [1, 2, 3]}}))
+
+        self.assertEqual(result['status'], 'ok')
+        self.assertEqual(result['data'], {'nested': [1, 2, 3]})
```
claravox: assertDictEqual?

claravox (approved):
LGTM!

PR #758, opened 2026-07-06
YDA-7082: fix inactive group notifications

stsnel:
Fix collection utility functions used by inactive
group notifications in order to resolve an issue where
data managers received notifications for groups that were
still active. The main change is that we need to take
into account leading zeroes when doing lexical comparisons
of timestamps. Also improve test coverage and error handling.


util/collection.py
```diff
@@ -299,3 +300,103 @@ def name_from_id(ctx: rule.Context, coll_id: str) -> str:
...
+        # This loop should only run once
+        for sub_row in iter_subcoll:
+            if int(sub_row[0]) > 0:
+                iter_subcoll.close()
+                return True
+
+    return False
+
+
+def has_collection_modify_date_after(ctx: rule.Context, collection_name: str, timestamp: int) -> bool:
```
claudisa-uu: I personally find the name of this function a bit confusing; would it make sense to call it `has_modify_date_after` without the `collection` part, since this is in collection.py anyway?

util/group.py
```diff
@@ -65,3 +65,18 @@ def get_category(ctx: 'rule.Context', grp: str) -> str | None:
     ret = ctx.uuGroupGetCategory(grp, '', '')
     x = ret['arguments'][1]
     return None if x == '' else x
+
+
+def get_list_research_groups(ctx: 'rule.Context') -> List[str]:
```
claudisa-uu: Would `get_research_groups_list` sound better?

claudisa-uu (approved):
Looks good to me!

PR #759, opened 2026-07-06
YDA-7089: fix statistics issue data last month

stsnel:
Export statistics previously sometimes had missing data for the latest month. This was because ruleset function get_resource_monthly_category_stats tries to step through monthly data by adding a month to the date iterator every iteration. This could result in the last month not being included in the monthly data if the current day of the month was before the day of the month of the first statistics data in the environment.


claudisa-uu (approved):
Looks good to me

PR #760, opened 2026-07-07
YDA-7087: improve group manager performance

stsnel:
Improve group manager performance for environments that have a significant number of groups without a group-level metadata schema.

Read information about configured category-level and environment-level metadata schemas in a dictionary, and then use this dictionary to perform lookups, rather than looking up this information repeatedly in the iCAT database for every group without a group-level schema.


claudisa-uu (approved):
Looks good to me

PR #761, opened 2026-07-14
YDA-6004 - Handle selection of multiple data measures in data request

kaur16:
YDA-6004 - Handle selection of multiple data measures in data request

### DIFF
--- a/datarequest.py
+++ b/datarequest.py
@@ -343,7 +343,7 @@ def api_upload_datarequest_data(ctx: rule.Context, path: str, data: Dict) -> api
     """
     if config.environment == 'development':
         try:
-            data_object.write(ctx, path, data)
+            jsonutil.write(ctx, path, data)
             return True
         except Exception:
             return False
@@ -403,6 +403,7 @@ def generate_request_id(ctx: rule.Context) -> int:
     for current_collection in collection.subcollections(ctx, coll, recursive=False):
         if str.isdigit(pathutil.basename(current_collection)) and int(pathutil.basename(current_collection)) > max_request_id:
             max_request_id = int(pathutil.basename(current_collection))
+    log.write(ctx, f"Max request id: {max_request_id}")
 
     return max_request_id + 1
 
@@ -1014,7 +1015,7 @@ def api_datarequest_submit(ctx: rule.Context, filename: str, request_id: str, dr
         return api.Error("permission_error", "Action not permitted.")
 
     # Construct data request collection and file path.
-    coll_path = "/{}/{}/{}".format(user.zone(ctx), DRCOLLECTION, request_id)
+    coll_path = "/{}/{}/{}".format(user.zone(ctx), DRCOLLECTION, req_id)
     file_path = "{}/{}".format(coll_path, DATAREQUEST + JSON_EXT)
 
     # If we're not working with a draft, initialize the data request collection
@@ -1075,10 +1076,10 @@ def api_datarequest_submit(ctx: rule.Context, filename: str, request_id: str, dr
 
     # If draft, set status
     if draft:
-        status_set(ctx, request_id, status.DRAFT)
+        status_set(ctx, req_id, status.DRAFT)
         # If new draft, return request ID of draft data request
         if not draft_request_id:
-            return {"requestId": request_id}
+            return {"requestId": req_id}
         # If update of existing draft, return nothing
         else:
             return
@@ -1100,13 +1101,13 @@ def api_datarequest_submit(ctx: rule.Context, filename: str, request_id: str, dr
 
     # Update data request status
     if data['datarequest']['purpose'] == "Analyses for data assessment only (results will not be published)":
-        status_set(ctx, request_id, status.DAO_SUBMITTED)
+        status_set(ctx, req_id, status.DAO_SUBMITTED)
     else:
         if data['datarequest']['attachments']['attachments'] == "Yes":
-            status_set(ctx, request_id, status.PENDING_ATTACHMENTS)
-            return {"pendingAttachments": True, "requestId": request_id}
+            status_set(ctx, req_id, status.PENDING_ATTACHMENTS)
+            return {"pendingAttachments": True, "requestId": req_id}
         else:
-            status_set(ctx, request_id, status.SUBMITTED)
+            status_set(ctx, req_id, status.SUBMITTED)
             return
 
 
--- a/datarequest.py
+++ b/datarequest.py
@@ -976,7 +976,6 @@ def api_datarequest_submit(ctx: rule.Context, request_id: str, draft: bool, draf
     """Persist a data request to disk.
 
     :param ctx:              Combined type of a callback and rei struct
-    :param filename:         Name of the file containing contents of the data request
     :param request_id:       Unique identifier of data request
     :param draft:            Boolean specifying whether the data request should be saved as draft
     :param draft_request_id: Unique identifier of the draft data request
--- a/datarequest.py
+++ b/datarequest.py
@@ -55,8 +55,8 @@ __all__ = ['api_datarequest_roles_get',
            'api_datarequest_signed_dta_path_get',
            'api_datarequest_data_ready',
            'rule_datarequest_review_period_expiration_check',
-           'api_generate_request_id',
-           'api_upload_datarequest_data']
+           'api_datarequest_generate_id',
+           'api_datarequest_upload_data']
 
 
 ###################################################
@@ -332,8 +332,9 @@ def available_documents_get(ctx: rule.Context, request_id: str, datarequest_type
 ###################################################
 
 @api.make()
-def api_upload_datarequest_data(ctx: rule.Context, path: str, data: Dict) -> api.Result:
-    """ Write datarequest data to path
+def api_datarequest_upload_data(ctx: rule.Context, path: str, data: Dict) -> api.Result:
+    """ Write datarequest data to path. This API endpoint ony works in development mode.
+    This API endpoint is only meant for testing.
 
     :param ctx:     Combined type of a callback and rei struct
     :param path:    Path to file containing datarequest data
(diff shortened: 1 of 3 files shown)


datarequest.py
```diff
@@ -349,14 +370,40 @@ def metadata_set(ctx: rule.Context, request_id: str, key: str, value: str) -> No
...
+    :returns: Data request ID
+    """
     coll           = "/{}/{}".format(user.zone(ctx), DRCOLLECTION)
     max_request_id = 0
 
     # Find highest request ID currently in use
     for current_collection in collection.subcollections(ctx, coll, recursive=False):
         if str.isdigit(pathutil.basename(current_collection)) and int(pathutil.basename(current_collection)) > max_request_id:
             max_request_id = int(pathutil.basename(current_collection))
+    log.write(ctx, f"Max request id: {max_request_id}")
```
stsnel: Is this line still needed, or was it a temporily debug statement?

datarequest.py
```diff
@@ -328,6 +331,24 @@ def available_documents_get(ctx: rule.Context, request_id: str, datarequest_type
...
 
+@api.make()
+def api_upload_datarequest_data(ctx: rule.Context, path: str, data: Dict) -> api.Result:
+    """ Write datarequest data to path
+
+    :param ctx:     Combined type of a callback and rei struct
+    :param path:    Path to file containing datarequest data
+    :param data:    Datarequest data
+
+    :returns:       Boolean - True if uploaded successfully else False
```
stsnel, line 343: I think it's best to mention in the docstring that this API endpoint only works in development mode, and is only meant for testing.

datarequest.py
```diff
@@ -328,6 +331,24 @@ def available_documents_get(ctx: rule.Context, request_id: str, datarequest_type
...
+    """
+    if config.environment == 'development':
```
stsnel, line 345: I think it's better to return an API error than to fail silently if the system is not in development mode.

datarequest.py
```diff
@@ -328,6 +331,24 @@ def available_documents_get(ctx: rule.Context, request_id: str, datarequest_type
...
+        try:
+            jsonutil.write(ctx, path, data)
+            return True
+        except Exception:
```
stsnel: I recommend logging the exception in case of failures. This makes it easier to debug API test failures.

datarequest.py
```diff
@@ -328,6 +331,24 @@ def available_documents_get(ctx: rule.Context, request_id: str, datarequest_type
 #                 Helper functions                #
 ###################################################
 
+@api.make()
+def api_upload_datarequest_data(ctx: rule.Context, path: str, data: Dict) -> api.Result:
```
stsnel: Can you adjust the function name so that it matches the name convention used by other API endpoints in this module? So something like *api_datarequest_upload_data* or similar?

datarequest.py
```diff
@@ -349,14 +370,40 @@ def metadata_set(ctx: rule.Context, request_id: str, key: str, value: str) -> No
     ctx.adminDatarequestActions()
 
 
+@api.make()
+def api_generate_request_id(ctx: rule.Context, draft_request_id: str) -> api.Result:
```
stsnel: Can you adjust the function name so that it matches the name convention used by other API endpoints in this module? So something like *api_datarequest_generate_id* or similar?

datarequest.py
```diff
@@ -1023,13 +1076,16 @@ def api_datarequest_submit(ctx: rule.Context, data: Dict, draft: bool, draft_req
...
         # If new draft, return request ID of draft data request
         if not draft_request_id:
-            return {"requestId": request_id}
+            return {"requestId": req_id}
         # If update of existing draft, return nothing
         else:
             return
+    else:
+        # Clean up the data file
+        data_object.remove(ctx, data_path)
```
stsnel: If removing the data object fails (e.g. because it is just being replicated), this could result in the API call failing, leaving the data request in an inconsistent state, without a status.

I think it'd be better to move the cleanup towards the end of the function (after the updating the status). And perhaps it can be wrapped in a try/except loop so that we can log the error but still continue with the request?

datarequest.py
```diff
@@ -922,16 +969,27 @@ def file_lock(ctx: rule.Context, coll_path: str, filename: str, readers: List[st
...
     """
+    # Read data from file
+    req_id = ""
+    if draft_request_id:
+        req_id = draft_request_id
+    else:
+        req_id = request_id
+
+    data_path = "/{}/{}/{}/{}".format(user.zone(ctx), DRCOLLECTION, req_id, filename)
+    data = jsonutil.read(ctx, data_path)
```
stsnel: Let's catch UUFileSizeError here (and fail if it is raised) in case the object exceeds the maximum size supported by jsonutil, to make it more robust.

datarequest.py
```diff
@@ -922,16 +969,27 @@ def file_lock(ctx: rule.Context, coll_path: str, filename: str, readers: List[st
...
     :returns: API status
     """
+    # Read data from file
+    req_id = ""
+    if draft_request_id:
+        req_id = draft_request_id
+    else:
+        req_id = request_id
+
+    data_path = "/{}/{}/{}/{}".format(user.zone(ctx), DRCOLLECTION, req_id, filename)
```
stsnel: I think we need to validate both `req_id` and `filename` here. Otherwise a user might be able to cause unintended behaviour by e.g. uploading a reserved filename (e.g. `datarequest.json`). Another option would be to just remove the `filename` parameter here and hard-code it.

datarequest.py
```diff
@@ -1023,13 +1076,16 @@ def api_datarequest_submit(ctx: rule.Context, data: Dict, draft: bool, draft_req
...
     if draft:
-        status_set(ctx, request_id, status.DRAFT)
+        status_set(ctx, req_id, status.DRAFT)
         # If new draft, return request ID of draft data request
         if not draft_request_id:
-            return {"requestId": request_id}
+            return {"requestId": req_id}
         # If update of existing draft, return nothing
         else:
             return
```
stsnel, line 1090: Should we not also clean up the uploaded datarequest object here, like in the non-draft branch?

tests/step_defs/api/test_api_datarequest.py
```diff
@@ -36,12 +42,40 @@ def api_datarequest_schema_get(user, schema_name):
...
+    return req_id
+
+
 @given('the Yoda datarequest submit API is queried with a data request to save as draft', target_fixture="api_response")
-def api_datarequest_save(user):
-    return api_request(
+def api_datarequest_save(user, req_id):
+    # Create datarequest collection and add permissions
+    # to upload datarequest file
+    api_request(
```
stsnel: I think it's best to check the results of this API request. If it fails, we'll want to fail immediately, so that it's easier to see where the failure started.

tests/step_defs/api/test_api_datarequest.py
```diff
@@ -36,12 +42,40 @@ def api_datarequest_schema_get(user, schema_name):
...
-        "datarequest_submit",
+        "datarequest_data_write_permission",
+        {
+            "request_id": req_id,
+            "action": "own"
+        }
+    )
+
+    # Upload datarequest file
+    api_request(
```
stsnel: I think it's best to check the results of this API request. If it fails, we'll want to fail immediately, so that it's easier to see where the failure started.

datarequest.py
```diff
@@ -1168,6 +1224,31 @@ def api_datarequest_attachment_post_upload_actions(ctx: rule.Context, request_id
...
+    :param ctx:        Combined type of a callback and rei struct
+    :param request_id: Unique identifier of the data request
+    :param action:     String specifying whether write permission must be granted ("grant") or
+                       revoked ("grantread" or "revoke")
+
+    :returns: None
+    """
+    # Create collection with request id if doesn't exist
+    # path as a parameter
+    datarequest_path = "/{}/{}/{}".format(user.zone(ctx), DRCOLLECTION, request_id)
```
stsnel, line 1251: We need validation for `request_id` here, since it's user provided data.

datarequest.py
```diff
@@ -1168,6 +1224,31 @@ def api_datarequest_attachment_post_upload_actions(ctx: rule.Context, request_id
...
+    if not collection.exists(ctx, datarequest_path):
+        collection.create(ctx, datarequest_path)
+
+    # Check if action is valid
+    if action not in ["grant", "grantread", "own"]:
+        return api.Error("InputError", "Invalid action input parameter.")
+
+    # Grant/revoke temporary write permissions
+    ctx.adminTempWritePermission(datarequest_path, action)
```
stsnel, line 1260: We need to check that the user is actually allowed to get this permission for this data request before changing permissions. Otherwise any user could just make themselves owner of any data request (including requests in which they dont have any role).

stsnel (commented):
The general structure looks good to me 👍 

I've added some comments about specific issues and questions. Can you also rebase the branch on development, please, so that it's easier to see the end result of this PR in combination with other bugfixes in the DR module?

datarequest.py
kaur16: The idea was that the data would be stored in datarequest-data.json until it is submitted.

kaur16: Added it in the end of the function instead.

datarequest.py
kaur16: Added the permissions check in yoda-portal side.

datarequest.py
kaur16: Removed

datarequest.py
kaur16: Done!

datarequest.py
kaur16: Done!

datarequest.py
kaur16: Done!

datarequest.py
kaur16: Moved the clean up at the end of the function and logged the error.

datarequest.py
kaur16: Remove the `filename` parameter

tests/step_defs/api/test_api_datarequest.py
kaur16: Added!

tests/step_defs/api/test_api_datarequest.py
kaur16: Added!

datarequest.py
kaur16: Added!

datarequest.py
stsnel: I'm concerned that this would still be a security issue, considering that users can connect to iRODS directly rather than via the portal. So they would still technically be able to give themselves write access to any data request.

PR #764, opened 2026-07-21
YDA-7210: improve performance get_groups_data

stsnel:
The get_groups_data function used to query all group metadata. However, this includes statistics data, which is irrelevant to the group manager and can be quite large on production environments (much larger than the group data itself). This could slow down the group manager considerably. Fix this by only querying for attributes that we actually need.


claudisa-uu (approved):
Looks good to me

PR #768, opened 2026-07-27
Datarequest: check result of datarequest_action_permitted

lwesterhof:


datarequest.py
```diff
@@ -1192,7 +1202,9 @@ def get_filename(file_path: str) -> str:
         return file_path.split('/')[-1]
 
     # Permission check
-    datarequest_action_permitted(ctx, request_id, ["PM", "DM", "DAC", "OWN"], None)
+    action_permitted = datarequest_action_permitted(ctx, request_id, ["PM", "DM", "DAC", "OWN"], None)
+    if not action_permitted:
+        return []
```
stsnel: It seems that returning a regular status here would hide the error condition from the frontend and user. Would it be better to return the api.Error instead? (e.g. `return action_permitted`)

stsnel (commented):
The changes look good to me in general. I have one question (see above)

stsnel (approved):
Nice, LGTM 👍 !

PR #769, opened 2026-07-27
YDA-7135: harden policies subfiles and object registration

stsnel:
Block adding subfiles, getting subfiles and file registration API
endpoints.


lwesterhof (approved):
LGTM!

PR #772, opened 2026-07-29
YDA-7141 - Modified SRAM to non-SRAM part of migration script to remove open invitations for internal users

kaur16:
YDA-7141 - Modified SRAM to non-SRAM part of migration script to remove open invitations for internal users


lwesterhof (approved):
LGTM!

PR #775, opened 2026-07-30
YDA-7142 - Added ability to remove invited metadata for internal user in external users sync script

kaur16:
YDA-7142 - Added ability to remove invited metadata for internal user in external users sync script


lwesterhof (approved):
LGTM

PR #777, opened 2026-08-05
YDA-7119: improve performance API call handling

stsnel:
Improve API call handling performance by switching to a higher performance JSON library (orjson).


claudisa-uu (approved):
Looks good to me! Maybe we could also optimize all other uses of jsonutil.parse() and jsonutil.dump() with the new fast alternatives that use orjson?

PR #778, opened 2026-08-05
YDA-7126: use atomic metadata operations for revision creation

lwesterhof:


revisions.py
```diff
@@ -625,22 +625,31 @@ def revision_create(ctx: rule.Context, print_verbose: bool, data_id: str, resour
...
-        avu.set_on_data(ctx, rev_path, constants.UUORGMETADATAPREFIX + "original_path", path)
-        avu.set_on_data(ctx, rev_path, constants.UUORGMETADATAPREFIX + "original_coll_name", parent)
-        avu.set_on_data(ctx, rev_path, constants.UUORGMETADATAPREFIX + "original_data_name", basename)
-        avu.set_on_data(ctx, rev_path, constants.UUORGMETADATAPREFIX + "original_data_owner_name", data_owner)
-        avu.set_on_data(ctx, rev_path, constants.UUORGMETADATAPREFIX + "original_coll_id", coll_id)
-        avu.set_on_data(ctx, rev_path, constants.UUORGMETADATAPREFIX + "original_modify_time", modify_time)
-        avu.set_on_data(ctx, rev_path, constants.UUORGMETADATAPREFIX + "original_group_name", group_name)
-        avu.set_on_data(ctx, rev_path, constants.UUORGMETADATAPREFIX + "original_filesize", data_size)
     except msi.Error as e:
         log.write(ctx, 'ERROR - The file could not be copied: {}'.format(str(e)))
```
stsnel, line 627: I think it would be best to do an early return here if the copy fails, since otherwise the function could return a success status if the copy failed halfway.

stsnel (commented):
I think this looks good in general 👍  I measured an approximately 2 x performance improvement for adding the revision AVUs on a dev environment.

I've added one question/comment regarding error handling (see below)

stsnel (approved):
LGTM 👍 !

PR #779, opened 2026-08-10
YDA-7169: fix publication related resource

stsnel:
Fix issue where publication flow failed on data packages with a related resource that had an identifier type but not identifier.


kaur16 (approved):
Looks good to me!

PR #780, opened 2026-08-10
YDA-7171: fix publication handling affiliations

stsnel:
Fix handling of affiliations without an affiliation name in the publication functions. Affiliations without a name are accepted by the current metadata schemas, so the publication functions should support this too.

### DIFF
--- a/json_datacite.py
+++ b/json_datacite.py
@@ -165,6 +165,31 @@ def get_funders(combi: dict) -> List:
     return funders
 
 
+def _process_affiliations_list(inputdata: List) -> List[dict]:
+    """Internal function for processing a list of affiliations
+
+    :param inputdata: List of affiliations in Yoda metadata format
+
+    :returns: List of dictionaries with affiliation data in DataCite format
+    """
+    affiliations: List[dict] = []
+    for aff in inputdata:
+        affiliation_data = {}
+        if isinstance(aff, dict):
+            if "Affiliation_Name" in aff and len(aff["Affiliation_Name"]):
+                affiliation_data["name"] = str(aff['Affiliation_Name'])
+            if "Affiliation_Identifier" in aff and len(aff["Affiliation_Identifier"]):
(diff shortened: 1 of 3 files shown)


json_datacite.py
```diff
@@ -165,6 +165,31 @@ def get_funders(combi: dict) -> List:
     return funders
 
 
+def _process_affiliations_list(input: List) -> List[dict]:
```
leonidastri: should we change input as it shadows the Python builtin?

json_datacite.py
```diff
@@ -165,6 +165,31 @@ def get_funders(combi: dict) -> List:
...
+    :param input: List of affiliations in Yoda metadata format
+
+    :returns: List of dictionaries with affiliation data in DataCite format
+    """
+    affiliations: List[dict] = []
+    for aff in input:
+        affiliation_data = {}
+        if isinstance(aff, dict):
+            if "Affiliation_Name" in aff and len(aff["Affiliation_Name"]):
+                affiliation_data["name"] = aff['Affiliation_Name']
```
leonidastri: in the old code we used to transform it to string. is it guaranteed that to be a string by the schema?

json_datacite.py
leonidastri: could there be an edge case causing a type error in this function, which could cause a crash on the publication?

leonidastri:
Clear implementation and easy to understand, great test coverage.  I have added some minor comments.

json_datacite.py
stsnel: Thanks, have re-added the str conversion to ensure who don't change behaviour.

json_datacite.py
stsnel: I've tried to incorporate the relevant edge cases in the unit tests. Do you have any suggestions for edge cases that still need to be covered?

leonidastri: I think the test coverage is good for now. We can always extend the test cases later.

leonidastri (approved):
LGTM!

PR #781, opened 2026-08-10
YDA-7172: fix publication function typing issues

stsnel:
Ensure that we convert publication status to a string when printing or returning it, so that we don't get type issues when we use it. This fixes (among other things) an issue where publishing a data packages that was published before fails.


leonidastri (approved):
LGTM!

PR #783, opened 2026-08-19
YDA-7090: refactor string formatting with f-strings and t-strings

claudisa-uu:
This PR replaces the old string.format() with f-strings and replaces GenQuery conditions that include collection names with t-strings to allow escaping of single quotes.

### DIFF
--- a/stats.py
+++ b/stats.py
@@ -11,6 +11,7 @@ from typing import List, Optional
 
 import genquery
 from dateutil.relativedelta import relativedelta
+from tstrings import t
 
 import groups
 from util import *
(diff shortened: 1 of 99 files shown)


stats.py
```diff
@@ -120,7 +120,7 @@ def api_resource_full_year_differentiated_group_storage(ctx: rule.Context, group
     total = []
     iter = genquery.row_iterator(
         "ORDER(META_USER_ATTR_NAME), META_USER_ATTR_VALUE",
-        "USER_NAME = '{}' AND META_USER_ATTR_NAME like '{}%%' AND USER_TYPE = 'rodsgroup'".format(group_name, constants.UUMETADATAGROUPSTORAGETOTALS),
+        f"USER_NAME = '{group_name}' AND META_USER_ATTR_NAME like '{constants.UUMETADATAGROUPSTORAGETOTALS}%%' AND USER_TYPE = 'rodsgroup'",
```
FuHsinyu: All the api calls were converted to f string, but this leads to potential injection. I tested this to get stats via bypassing the role permission, and I succeed. What I did was to use console to call with e.g, "research-<group>' || like '%" 

I suggest to use t string here as well (for all api requests), and use GenQuery2? (maybe good to have them in a separate PR)

FuHsinyu (changes requested):
The fix is very substantial and useful, see comments for issues about potential injection and bug

stats.py
FuHsinyu: A follow-up ticket is created at https://utrechtuniversity.atlassian.net/browse/YDA-7216

PR #784, opened 2026-08-19
YDA-7193: fix DR module overview

stsnel:
The data request module overview wasn't working
correctly after introduction of type checking,
because the portal was passing API parameters
as strings rather than booleans (and the ruleset
function type annotation was inconsistent with how it evaluated parameters).


claudisa-uu (approved):
Looks good to me

